PR #12, opened 2018-01-19
chore: prettier/eslint config with git hook

enguerran:
I updated `.eslintrc.json` according to https://prettier.io/docs/en/eslint.html#turn-off-eslint-s-formatting-rules that _Turn off ESLint's formatting rules_.
So:

- eslint is used to check code quality
- prettier is used to format the code

And prettier will run on every commit with https://prettier.io/docs/en/precommit.html#option-2-pretty-quick-https-githubcom-azz-pretty-quick


ptbrowne:
Does it make sense to update eslint-plugin-cozy-app with your changes ?

PR #26, opened 2018-01-29
feat: enhance pick-contact modal

enguerran:
- add a cancel button
- add a global layout
- add a selection number
- add a header

In this PR I will:

- [x] add a layout to the modale
- [x] add a counter for selected contacts
- [x] i18nize strings
- [ ] extract css rules and component for cozy-ui
- [ ] import css rules and component from cozy-ui

I don't know if I will do the two last points, we'll see.

### DIFF
--- a/src/styles/intent.styl
+++ b/src/styles/intent.styl
@@ -1,0 +1,52 @@
+@require 'settings/palette.styl'
+.app-wrapper
+    width 100%
+.app-content
+    height 100vh
+
+.intent-layout
+    height 100%
+    display flex
+    align-items stretch
+    flex-direction column
+
+
+.intent-footer
+    height 4.5rem
+    background-color paleGrey
+    box-shadow inset 0 -1px 0 0 silver
+    display flex
+    align-items center
+    padding .5rem 1rem
+
+    .intent-footer-label
+        flex 1
+
+.intent-main
+    flex 1
+    overflow-y scroll
+    width 100%
+
+.intent-header
+    height 2rem
+    width 100%
+    background-color paleGrey
+    box-shadow inset 0 -1px 0 0 silver
+    padding 1rem
+    display flex
+
+    .intent-header-title
+        display flex
+        margin 0
+        align-items center
+        font-size 1.5rem
+        font-weight normal
+        color charcoalGrey
+
+        & span
+            margin-right .25rem
+        & strong
+            font-weight bold
+
+    .intent-header-icon
+        margin-right .45rem
--- a/src/components/IntentApp.jsx
+++ b/src/components/IntentApp.jsx
@@ -1,70 +1,27 @@
-/* global cozy */
 import React from "react";
-import PropTypes from "prop-types";
-import { Button } from "cozy-ui/react";
 
-import ContactsList from "./ContactsList/ContactsList";
-import { withContacts } from "./ContactsList";
+import "styles/intent";
 
-const ConnectedContactsList = withContacts(ContactsList);
+import PickContacts from "components/PickContacts";
 
-class PickContacts extends React.Component {
-  state = {
-    service: null,
-    selection: []
-  };
-  async componentDidMount() {
-    cozy.client.intents
-      .createService(this.props.intentId, window)
-      .then(service => {
-        this.setState(state => ({ ...state, service }));
-      });
-  }
-
-  pickContacts = async () => {
-    if (this.state.service) {
-      try {
-        this.state.service.terminate({ contacts: this.state.selection });
-      } catch (error) {
-        this.state.service.throw(error);
-      }
-    }
-  };
-
-  toggleSelection = data => {
-    const index = this.state.selection.indexOf(data);
-    this.setState(state => ({
-      ...state,
-      selection:
-        index === -1
-          ? [...state.selection, data]
-          : [
-              ...state.selection.slice(0, index),
-              ...state.selection.slice(index + 1)
-            ]
-    }));
-  };
-
-  render() {
-    return (
-      <div>
-        <Button onClick={this.pickContacts}>Select</Button>
-        <ConnectedContactsList
-          selection={this.state.selection}
-          onSelect={this.toggleSelection}
-        />
-      </div>
-    );
-  }
-}
-PickContacts.propTypes = {
-  intentId: PropTypes.string.isRequired
+const arrToObj = (obj = {}, varval = ["var", "val"]) => {
+  obj[varval[0]] = varval[1];
+  return obj;
 };
 
+const getQueryParameter = () =>
+  window.location.search
+    .substring(1)
+    .split("&")
+    .map(varval => varval.split("="))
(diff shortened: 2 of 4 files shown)


enguerran:
[(3) ETD je peux faire sélectionner des contacts en inter-app](https://trello.com/c/1cXmqQAC/723-3-etd-je-peux-faire-s%C3%A9lectionner-des-contacts-en-inter-app)

enguerran:
Here is the final shot



What do you think?

enguerran:
Do not hesitate to be more specific with the CSS problems. It could help understand, improve and correct.

src/styles/intent.styl
```diff
@@ -0,0 +1,52 @@
+@require 'settings/palette.styl'
+.app-wrapper
+    width 100%
+.app-content
+    height 100vh
```
y-lohse: I'm always cautious with `vh`. Unless there's a clear benefit of using it instead of `100%`, using a percent value is safer.

src/styles/intent.styl
```diff
@@ -0,0 +1,52 @@
...
+    height 100%
+    display flex
+    align-items stretch
+    flex-direction column
+
+
+.intent-footer
+    height 4.5rem
+    background-color paleGrey
+    box-shadow inset 0 -1px 0 0 silver
```
y-lohse: Since we have 2 identical shadows, we should use a variable for this. I also believe our shadows are standardized and there are predefined ones in cozy-ui.

src/styles/intent.styl
```diff
@@ -0,0 +1,52 @@
...
+    display flex
+    align-items center
+    padding .5rem 1rem
+
+    .intent-footer-label
+        flex 1
```
y-lohse: I think `flex-grow 1` is more explicit than `flex 1` and it doesn't change the `flex-basis`, which is what you intend most of the time.

src/styles/intent.styl
```diff
@@ -0,0 +1,52 @@
...
+
+.intent-main
+    flex 1
+    overflow-y scroll
+    width 100%
```
y-lohse: Are the width here and in `intent-header` really useful? Since their parent has `align-items stretch` I think they'll automatically have 100% width.

src/styles/intent.styl
enguerran: Ok, I don't know the difference between `100%` and `100vh` or `100vw`. I saw unpleasant behavior with the classical `100%` so I tried `100vh` that works for my needs.

src/styles/intent.styl
enguerran: I did not know that, thanks for the tips. Nevertheless I checked [cozy-ui docs](http://cozy.github.io/cozy-ui/styleguide) and I found nothing about shadows.
I cargo-cult this line from https://github.com/cozy/cozy-collect/blob/master/src/styles/services.styl#L51

src/styles/intent.styl
enguerran: :+1: but too late :/

src/styles/intent.styl
enguerran: Not sure, I cargo-cult what I found here https://jsfiddle.net/srbakfzw/
And I believe if I removed it, it breaks.

I try to generalize the concept in a codesandbox app you can find here https://codesandbox.io/s/lx4z0yvz9m

enguerran:
> I've added my css-related comments, but I'm not sure about all of them — normally I would have double-checked before making them. So take it with a grain of salt :)

That is the exact purpose of github's Pull Request: to communicate about the code. Your review will never be taken as god's speach, but your help and your opinion are always valuable. Do not hesitate to raise up any doubt, or to express your opinion. We are discussing about code, not the PR's author. Be my guess to give hints, advices, opinions or any useful remarks about the code.

src/styles/intent.styl
y-lohse: So `height: 100vh` is `100% of the viewport height`, ie. the whole window. `height: 100%` is `100% of the parent's height`, which can be a lot of different things.
But on iOS at least, `vh` seems unreliable. i haven't investigated too much, but:

> vh on iOS is reported to include the height of the bottom toolbar in the height calculation, and the width of the sidebar (bookmarks) in the vw width calculation.

https://caniuse.com/#feat=viewport-units

Most of the time, when you want an element to take `100%` of the height, it's parent already uses `100%` of the height, so using `%` instead of `vh` is safer.

src/styles/intent.styl
y-lohse: Looks like I'm wrong. i'm pretty sure there *used to be* box shadow variables, but I can't find them anymore either..?

src/styles/intent.styl
y-lohse: [Here's a fork of your sandbox](https://codesandbox.io/s/m5zln0rzwy) without the `width 100%` which seems to work the same across all major browsers :)

ptbrowne: @y-lohse they have automatic height since the flex-direction is column. width is not impacted by this.

y-lohse: Erm, no, I don't think so?

Since `flex-direction` is set to `column`, the main axis is vertical. So `justify-content` affects the *main axis*, ie. the `height`, and `align-items` affects the cross-axis, so the `width`.
So setting `align-items stretch` on a flexbox with `flex-direction column` *does* affect the children's width.

src/styles/intent.styl
GoOz: Not just iOS, Chrome mobile does this too. They kinda interpreted the W3C Specification differently than others… The "bug" was reported a lot and they said this was exactly what they wanted and had no intention what so ever to change this. So yeah, `vh` is a doomed unit but `vw` is cool though.

src/styles/intent.styl
GoOz: Not to my knowledge… maybe in the first experimental shot of Cozy-UI. It rings a bell somehow…

src/styles/intent.styl
GoOz: also, is the nesting really relevant? The naming should be enough to figure out its parent. No need for a nesting that add nothing but heavier selectors, I think.

bmillemathias:
Coming back on the design of the pick-contact window, that's would be useful to have a search box with filter as you type to quickly find contacts

jsilvestre:
Hi @bmillemathias,

You are not off-topic, but even if you where that'd be okay. We will never ask you to fit our development process (most of the product/design  process is done in Trello, not Github) so feel free to comment any time you want, on any topic you want, if it seems somewhat relevant to you, we'll handle the feedback!

Regarding the feedback itself, it's a early version of the modal we got in the product early so a partner can integrate it in its app, which is why it's not perfect. It'll be the screen that any app will be able to use to select contacts. As a result, a "quick search" is absolutely relevant on this screen. It's in the backlog, but I can't promise it will end up there in the first release.

Thank you for the feedback!
Joseph.

PR #38, opened 2018-02-06
chore: don't lint before tests

y-lohse:
While writing tests, I found the linting before the tests quite annoying. The linter will complain about `console.log` statements and other not too important things, and refuse to run the tests until they are fixed.

I also don't think there's a lot of value in linting before testing. They are just 2 different things. What do you think?


enguerran:
I think you're right and I think you should add the lint task to travis https://github.com/cozy/cozy-contacts/blob/master/.travis.yml#L31-L33

y-lohse:
Ah yes of course, thanks!

PR #63, opened 2018-02-23
fix: disallow creating a contact without name

y-lohse:
It starts to smell when we have to drill down the props like that, doesn't it ?

### DIFF
--- a/src/components/ContactCard/ContactForm.jsx
+++ b/src/components/ContactCard/ContactForm.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Form } from "react-final-form";
-import ContactFieldForm from "./ContactFieldForm";
+import ContactFormField from "./ContactFormField";
 import { Button } from "cozy-ui/react/Button";
 import { translate } from "cozy-ui/react/I18n";
 
@@ -22,7 +22,8 @@ const fields = [
   {
     name: "familyName",
     icon: null,
-    type: "text"
+    type: "text",
+    required: true
   },
   {
     name: "phone",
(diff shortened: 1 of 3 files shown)


y-lohse:
[(3) Créer un nouveau contact simple](https://trello.com/c/bZaQ18TS/716-3-cr%C3%A9er-un-nouveau-contact-simple)

y-lohse:
Well, `ContactFieldForm` and `ContactFieldInput` have nearly the same list of props. `ContactFieldForm` just passes most of them along to `ContactFieldInput` and I'm not sure it's a good thing or how things could be done differently.

goldoraf:
@y-lohse I see. Well, first I'd rename `ContactFieldForm` to `ContactField`, the `-Form` suffix makes no sense to me (If anything, I'd accept `ContactFormField`). Second, why not rewrite it like this:
```jsx
const ContactField = ({
  icon,
  label,
  ...inputProps
}) => (
  <div className="contact-form__field">
    <label className="contact-form__label">
      {icon && (
        <Icon icon={icon} color="coolGrey" className="contact-form__icon" />
      )}
      {label}
    </label>
    <ContactFieldInput
      {...inputProps}
    />
  </div>
);
```
i.e., we extract the `Field` props from the overall props and pass the rest to the `Input` child. What do you think of this pattern?

y-lohse:
Mmh, I think it looks cleaner, but also less explicit :/ And the PropTypes still show the problem.

I think I'll leave it like this for now, this part is likely to change again when we support multiple emails / phones / etc. That will be the right time to reorganize this. Thanks for the insight!

y-lohse:
[(3) Créer un nouveau contact simple](https://trello.com/c/bZaQ18TS/716-3-cr%C3%A9er-un-nouveau-contact-simple)

PR #67, opened 2018-02-26
upgrade cozy-ui

y-lohse:
Upgrades to the lastest version of cozy-ui and fixes a couple of things that broke along the way.

### DIFF
--- a/src/components/ContactCard/ContactIdentity.jsx
+++ b/src/components/ContactCard/ContactIdentity.jsx
@@ -2,20 +2,11 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Avatar } from "cozy-ui/react/Avatar";
 import contactPropTypes from "../ContactPropTypes";
-
-const getFullContactName = name =>
-  ["namePrefix", "givenName", "additionalName", "familyName", "nameSuffix"]
-    .map(part => name[part])
-    .join(" ")
-    .trim();
+import { getFullContactName, getInitials } from "../../helpers/contacts";
 
 const ContactIdentity = ({ name, groups }) => (
   <div className="contact-card-identity">
-    <Avatar
-      firstname={name.givenName}
-      lastname={name.familyName}
-      size="medium"
-    />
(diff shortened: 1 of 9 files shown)


src/components/ContactCard/ContactIdentity.jsx
```diff
@@ -2,20 +2,11 @@ import React from "react";
 import PropTypes from "prop-types";
 import { Avatar } from "cozy-ui/react/Avatar";
 import contactPropTypes from "../ContactPropTypes";
-
-const getFullContactName = name =>
-  ["namePrefix", "givenName", "additionalName", "familyName", "nameSuffix"]
-    .map(part => name[part])
-    .join(" ")
-    .trim();
+import { getFullContactName, getInitials } from "../../helpers/contacts";
```
goldoraf: I'm not at ease with these helper functions in another file, I think we're losing some kind of consistency. What do you think of using the `ViewModel` pattern here (sometimes called `Presenter` pattern in the Rails world), that is: having a class that wrap a raw `io.cozy.contacts` object and that handles view logic like sorting the fields, generating the initials, etc.
`ContactCard` would become something like this: 
```jsx
const ContactCard = ({ title, contact, renderActions }) => {
  const viewModel = new ContactViewModel(contact)
  return (
    <div>
      <header className="contact-card-header">
        <ContactIdentity initials={viewModel.getInitials()} fullname={viewModel.fullname()} groups={[]} />
        {renderActions()}
      </header>
      <ContactFields fields=viewModel.getOrderedFields()} title={title} />
    </div>
  );
};
```

y-lohse: Well, i don't know, what would be the advantages of using such a `ViewModel` vs pure, standalone functions ? I have nothing against it per se, but I don't see why it would be compelling either.

y-lohse:
[(3) Créer un nouveau contact simple](https://trello.com/c/bZaQ18TS/716-3-cr%C3%A9er-un-nouveau-contact-simple)

PR #78, opened 2018-03-07
Multiple emails/phones/addresses

y-lohse:
I've added comments where needed, but the changes are actually quite small. I've "only" implemented [`FieldArray`](https://github.com/final-form/react-final-form#field-arrays) from final-form.


y-lohse:
[(3) Créer un contact avec plusieurs champs d'un même type](https://trello.com/c/863mfSA9/796-3-cr%C3%A9er-un-contact-avec-plusieurs-champs-dun-m%C3%AAme-type)

src/components/ContactCard/ContactForm.jsx
```diff
@@ -66,6 +70,11 @@ const fields = [
   }
 ];
 
+const initialFieldValues = fields.reduce((initialValues, { name, isArray }) => {
+  initialValues[name] = isArray ? [undefined] : undefined;
+  return initialValues;
+}, {});
+
```
y-lohse: We are now using final-form's `FieldArray`, and arrays of field start empty by default. So we must init those fields as an array containing one element.

src/components/ContactCard/ContactForm.jsx
```diff
@@ -136,24 +135,34 @@ class ContactForm extends React.Component {
...
+                      isArray={isArray}
+                      renderInput={inputName => (
+                        <ContactFieldInput
+                          name={inputName}
+                          type={type}
+                          placeholder={t(`placeholder.${name}`)}
+                          required={required}
+                          withLabel={hasLabel}
+                          labelPlaceholder={t("placeholder.label")}
+                        />
```
y-lohse: I've changed this, I think it's better but still not pefect.

The structure is always the same

```
< ContactFormField>
  < ContactFieldInput />
</ContactFormField>
```

But `ContactFieldInput` needs a lot of props that are unrelated to `ContactFormField`.

Before, we passed a lot of props to `ContactFormField`, who then used a few, and passed the rest down to `ContactFieldInput`. By using a renderProp, we can give each component only the props it really needs to do it's job.

What I still don't like too much is that we could render *anything* in `ContactFormField.renderInput`, when in reality, we really expect a `ContactFieldInput`, or at least a `Field` from final-form.

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -92,32 +91,57 @@ const ContactFormField = ({
...
+                    <button
+                      type="button"
+                      onClick={() => fields.push(undefined)}
+                      className="contact-form__meta-button contact-form__meta-button--add"
+                    >
+                      <Icon icon={IconPlus} />
+                    </button>
+                  )}
+                </div>
+              );
```
y-lohse: I find the contact of the `fields.map` function a bit hard to read, but I don't think we would gain much from splitting it elsewhere.

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -75,49 +92,90 @@ ContactFieldInput.defaultProps = {
...
+      </div>
+    );
+  }
+
+  addField = fields => fields.push(undefined);
+
+  removeField = (fields, index) => {
+    const isLastRemainignField = fields.length === 1;
+    fields.remove(index);
+    if (isLastRemainignField) this.addField(fields);
```
y-lohse: Yes, it'd be easier to do a `if (fields.length === 0)` after the `remove()`, but `remove()` seems to be async or something, because immediately afterwards `fields.length` is still `1`.

src/components/ContactCard/ContactForm.jsx
edas: [Style] It seems weird to do it that way, but I don't have something better to propose. However, if usefull: Shouldn't you put your comment directly in the code instead of a github comment ?

src/components/ContactCard/ContactForm.jsx
edas: The new way seems ok to me. The renderInput being restricted is not a real problem for me. A line in the documentation should do the trick, and the code will fail if ever someone tries to do something else.

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -32,17 +44,20 @@ class ContactFieldInput extends React.Component {
...
     } = this.props;
-    const { renderLabel } = this.state;
+    const { renderLabel, hasFocus } = this.state;
+    const focusedClass = hasFocus
+      ? " contact-form__input-wrapper--focused"
+      : "";
 
     return (
-      <div className="contact-form__input-wrapper">
+      <div className={`contact-form__input-wrapper${focusedClass}`}>
```
edas: [Not blocking] Not fond of this. I recommend https://github.com/JedWatson/classnames , even if the logic is very basic

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -65,8 +82,8 @@ ContactFieldInput.propTypes = {
   type: PropTypes.string.isRequired,
   placeholder: PropTypes.string,
   withLabel: PropTypes.bool,
-  required: PropTypes.bool,
-  labelPlaceholder: PropTypes.string
+  labelPlaceholder: PropTypes.string,
+  required: PropTypes.bool
```
edas: [Meta] When you review your changes, it's best if you can revert such unneeded changes before commit.

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -75,49 +92,90 @@ ContactFieldInput.defaultProps = {
...
 ContactFormField.defaultProps = {
   icon: null,
-  inputWithLabel: false,
-  required: false,
-  placeholder: "",
-  labelPlaceholder: ""
+  isArray: false
 };
-
-export default ContactFormField;
```
edas: [Style] Le fichier finit par être long, concerne deux composants distincts (même s'ils sont liés), est-ce que ça ne mériterait pas de faire deux fichiers, un par composant ? (quitte à ce que celui-ci importe ContactFieldInput et le ré-exporte, pour garder la compatibilité)

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -75,49 +92,90 @@ ContactFieldInput.defaultProps = {
...
+        <label className="contact-form__label">
+          {icon && (
+            <Icon icon={icon} color="coolGrey" className="contact-form__icon" />
+          )}
+          {label}
+        </label>
+        {isArray ? (
+          <FieldArray name={name}>
+            {({ fields }) => (
+              <div className="contact-form__inputs-wrapper">
```
edas: [Not blocking] à vue de nez, on ne pourrait pas le mettre en parent de FieldArray plutôt que l'inverse ? ça permettrait de le mutualiser avec la L123 https://github.com/cozy/cozy-contacts/pull/78/files#diff-7fe9fa8d96002866bee0233c53e45554R123

src/components/ContactCard/ContactForm.jsx
y-lohse: True, the line is confusing enough that it warrants an explanation in the code.

src/components/ContactCard/ContactFormField.jsx
y-lohse: Oui en effet, je regarderais de plus près mais je pense que les 2 composants sont suffisamment distincts à présent pour les séparer.

src/components/ContactCard/ContactFormField.jsx
ptbrowne: +1. Using git add -p helps here. If you are on mac, gitup is a very simple GUI to stage hunks and reverting hunks.

src/components/ContactCard/ContactFormField.jsx
y-lohse: Malheureusement non.
Si on enlève ce `div` ici, l'enfant-fonction de `FieldArray` va renvoyer un tableau de composants (le résultat de `fields.map`). Mais avoir un composant qui renvois un tableau de composant est une nouvelle feature de react 16, que preact ne supporte pas (encore). Donc dans notre cas, il faut renvoyer un seul composant… ici le `div` wrapper.

src/components/ContactCard/ContactFormField.jsx
```diff
@@ -1,25 +1,37 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { Field } from "react-final-form";
+import { FieldArray } from "react-final-form-arrays";
```
ptbrowne: I am not aware of `react-final-form-arrays`, it would be cool if you could present it at a crosspoll :)

package.json
```diff
@@ -59,14 +59,17 @@
     "stylint": "1.5.9"
   },
   "dependencies": {
+    "classnames": "^2.2.5",
```
enguerran: pin the version

package.json
```diff
@@ -59,14 +59,17 @@
...
     "cozy-bar": "4.8.9",
     "cozy-client": "1.0.0-beta.1",
     "cozy-scripts": "0.5.4",
     "cozy-stack-link": "1.0.0-beta.1",
     "cozy-ui": "7.5.0",
     "final-form": "4.2.1",
+    "final-form-arrays": "^1.0.4",
```
enguerran: pin the version

package.json
```diff
@@ -59,14 +59,17 @@
...
     "preact": "8.2.7",
     "preact-compat": "3.18.0",
-    "react-final-form": "3.1.2"
+    "react-final-form": "3.1.2",
+    "react-final-form-arrays": "^1.0.4"
```
enguerran: pin the version

src/assets/icons/small-cross.svg
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="8px" height="8px" viewBox="0 0 8 8" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 49 (51002) - http://www.bohemiancoding.com/sketch -->
+    <title>icons/16/cross</title>
+    <desc>Created with Sketch.</desc>
+    <defs>
+        <path d="M4,2.58578644 L6.29289322,0.292893219 C6.68341751,-0.0976310729 7.31658249,-0.0976310729 7.70710678,0.292893219 C8.09763107,0.683417511 8.09763107,1.31658249 7.70710678,1.70710678 L5.41421356,4 L7.70710678,6.29289322 C8.09763107,6.68341751 8.09763107,7.31658249 7.70710678,7.70710678 C7.31658249,8.09763107 6.68341751,8.09763107 6.29289322,7.70710678 L4,5.41421356 L1.70710678,7.70710678 C1.31658249,8.09763107 0.683417511,8.09763107 0.292893219,7.70710678 C-0.0976310729,7.31658249 -0.0976310729,6.68341751 0.292893219,6.29289322 L2.58578644,4 L0.292893219,1.70710678 C-0.0976310729,1.31658249 -0.0976310729,0.683417511 0.292893219,0.292893219 C0.683417511,-0.0976310729 1.31658249,-0.0976310729 1.70710678,0.292893219 L4,2.58578644 Z" id="path-1"></path>
+    </defs>
+    <g id="📦-symbols" stroke-width="1">
```
enguerran:
```
id="📦-symbols"
```
that's weird!

By the way `small-cross.svg` and `small-plus.svg` got the same `id` on the `<g />` tag. It may causes bugs, don't it?

src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -0,0 +1,93 @@
...
+  state = {
+    renderLabel: false,
+    hasFocus: false
+  };
+
+  onFocus = () => {
+    this.setState({
+      renderLabel: this.props.withLabel,
+      hasFocus: true
+    });
```
enguerran: Do you forget to use functional form of `setState`?

src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -0,0 +1,93 @@
...
+import PropTypes from "prop-types";
+import { Field } from "react-final-form";
+import classnames from "classnames";
+
+const getInputComponent = inputType =>
+  inputType === "textarea" ? "textarea" : "input";
+
+class ContactFieldInput extends React.Component {
+  state = {
+    renderLabel: false,
```
enguerran: As a boolean, this variable sould be named with a prefix like `is` or `has`, should not it?

src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -0,0 +1,93 @@
...
+      renderLabel: e.target.value && this.props.withLabel,
+      hasFocus: false
+    });
+  };
+
+  onSecondaryInputBlur = () => {
+    this.setState({
+      hasFocus: false
+    });
+  };
```
enguerran: I don't know what this PR aims to do. But at first sight, this component is a bit too complex. I can conclude on that because of these two functions `onMainInputBlur` and `onSecondaryInputBlur`. I keep trying to understand what the `ContactFieldInput` component is responsible for.

src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+  render() {
+    const {
+      name,
+      type,
+      placeholder,
+      required,
+      withLabel,
+      labelPlaceholder
```
enguerran: I don't get what is `placeholder` and what is `labelPlaceholder`.

src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -0,0 +1,93 @@
...
+              className="contact-form__label-input"
+              placeholder={labelPlaceholder}
+              onFocus={this.onFocus}
+              onBlur={this.onSecondaryInputBlur}
+            />
+          )}
+      </div>
+    );
+  }
+}
```
enguerran: Regarding the `render` function, I get what you are trying to do. This component named `ContactFieldInput` is not responsible for any input nor any field, but it render a `<Field />` component and another conditional `<Field />` component.
This is not straight forward to understand if the second field will be displayed or not, first because it lays on two boolean `withLabel && renderLabel`, second because it takes a lot of props and distribute those props into two unspecified `<Fields />`.

I recommend:
- to functionally name both underlying components with a semantic name
- to factorize under a named function both boolean with a name that is not `shouldRenderSecondaryField` of course

Those recommendations are not mandatory, just a matter of what could be done to help developers understand what is going on.

src/components/ContactCard/ContactForm.jsx
enguerran: Ouais, c'est vraiment weird!

enguerran:
Ah oui, ce serait bien de fixer les versions des packages ajoutés par la PR.

y-lohse:
Je laisserais renovate passer dessus pour fixer les versions, si je ne dit pas de betises ca a aussi un impact sur le `yarn.lock`. Une machine fera ca mieux que moi :)

src/components/ContactCard/ContactFieldInput.jsx
y-lohse: I'm leaving this as it is for now, but I agree with your sentiment.

To clarify, some of the fields must be able to split themselves into 2 fields, one for the actual value, and one to attach a label to the value.
I'm not sure it would help to create standalone components for them, as they are just a `<Field />`. But the amount of props the parent component receives, and the similar names (`placeholder`, `labelPlaceholder`, etc), are not helping.

I think when we work on editing contacts, we'll have another shot at making this better.

edas: I am not sure about the "I agree but won't do it" :-)  
If you agree, do it!

Renaming things may not be very hard, and if it's needed to achieve a good quality and you both agree on it, we should certainly do it and not leave it for tomorow

PR #88, opened 2018-03-16
chore: migrate to latest cozy-ui

y-lohse:
I've migrated the buttons and stuff to the latest cozy-ui, but we have a problem with the way Modals work now.

So far, we have decided that `ContactCard` and `ContactForm` are simply the *content* of a modal, but they could just as well be rendered inside a standalone page. So they are totally decoupled from the Modal API.

capture d ecran 2018-03-09 16 56 14

This is good on paper, but a problem in practice.

With the new API, it is quite important that we start using the `ModalHeader` and `ModalFooter` components in order to get the proper styling and scrolling behavior.

So for example [here](https://github.com/cozy/cozy-contacts/blob/master/src/components/ContactCard/ContactCard.jsx#L27), we would have to extract the `header` part and move it to [here, maybe](https://github.com/cozy/cozy-contacts/blob/master/src/components/Modals/ContactCardModal.jsx#L38). And something similar would happen with the ContactForm.

Does this sounds reasonable? Any alternatives?

cc @enguerran @goldoraf @GoOz

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "cozy-client": "1.0.0-beta.1",
     "cozy-scripts": "0.5.5",
     "cozy-stack-link": "1.0.0-beta.1",
-    "cozy-ui": "7.7.0",
+    "cozy-ui": "7.7.3",
     "final-form": "4.2.1",
     "final-form-arrays": "1.0.4",
     "preact": "8.2.7",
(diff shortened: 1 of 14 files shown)


y-lohse:
I've updated the `ContactCard` / `ContactModal` relationship as discussed friday in [this commit](https://github.com/cozy/cozy-contacts/pull/88/commits/8ba7a82260f2e3fb92d78ca9a3f0007c2e100034) !

Everything should be good to go now.

y-lohse:
[(5) Gérer les groupes dans la fiche de contact](https://trello.com/c/NiAXu1HW/729-5-g%C3%A9rer-les-groupes-dans-la-fiche-de-contact)

edas:
👍 for me

PR #99, opened 2018-03-23
fix: keyboard interactions on group manager

y-lohse:
- prevent the tab key from selecting values
- capture the "escape" key propagation when the component has focus so the modal doesn't get closed


y-lohse:
[(5) Gérer les groupes dans la fiche de contact](https://trello.com/c/NiAXu1HW/729-5-g%C3%A9rer-les-groupes-dans-la-fiche-de-contact)

PR #104, opened 2018-03-27
Confirmation modal before deleting a contact

y-lohse:
The `ModalCard` is now a class with a state, to decide whether the confirmation modal is open or not. And after the deletion, we show an alert message. The changes are pretty straightforward!

Just waiting on the cozy-ui release.

### DIFF
--- a/src/components/Modals/ContactCardModal.jsx
+++ b/src/components/Modals/ContactCardModal.jsx
@@ -79,12 +79,12 @@ class ContactCardModal extends React.Component {
         {showConfirmDeleteModal && (
           <Modal
             into="body"
-            title="Supprimer ce contact ?"
-            description="Apu contact"
-            primaryText="ouais"
+            title={t("delete-confirmation.title")}
+            description={t("delete-confirmation.description")}
+            primaryText={t("delete")}
             primaryType="danger"
             primaryAction={this.deleteContact}
-            secondaryText="en fait non"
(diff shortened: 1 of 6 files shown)


y-lohse:
[(1) Ajouter une modale de confirmation lors de la suppression d'un contact](https://trello.com/c/pnIW5A5u/823-1-ajouter-une-modale-de-confirmation-lors-de-la-suppression-dun-contact)

src/components/Modals/ContactCardModal.jsx
```diff
@@ -35,29 +35,65 @@ ContactCardMenu.propTypes = {
...
+                deleteAction={{
+                  label: t("delete"),
+                  action: this.toggleConfirmDeleteModal
+                }}
+              />
+            </ModalHeader>
+          )}
+          renderBody={children => <ModalContent>{children}</ModalContent>}
+        />
+        {showConfirmDeleteModal && (
```
enguerran: If `isConfirmDelete` then render `<Modal />`. I think naming a Boolean as a condition or a question is easier to read: _isSomethingGood_, _shouldBeGood_, etc. What do you think?

y-lohse: In my head, I read it as *show confirm delete modal ?*, coupled with the fact that it's clearly a boolean. But I don't mind changing it if it helps other people!

PR #109, opened 2018-03-29
feat: sort contact by familyName

enguerran:
:tada:


enguerran:
Yeah, it's exported, and then I'm thinking we'll have the option earlier to choose the sort order. But you're right, the git history is there to cover us.

PR #155, opened 2018-05-23
Create contact intent

y-lohse:
Adds a second intent that uses the `ContactForm` component to allow creating a contact through an intent. 
If the intent data has a flag `me: true`, we add that information to the contact's metadata. This contact becomes the assumed profile for the cozy's owner and get's a little "me" 🎏 in the list.

Since we now have 2 intents in this app, i've restructured things slightly, using a technique borrowed from cozy-store.


src/components/ContactCard/ContactForm.jsx
```diff
@@ -6,6 +6,7 @@ import ContactFormField from "./ContactFormField";
 import ContactFieldInput from "./ContactFieldInput";
```
y-lohse: A little extra with this PR: we want to render the buttons for the creation form… elsewhere. Generally in a fixed modal footer.

I've used portals to make that possible, and successfully used it in the intent. However, when using this with the `ContactFormModal`, we run into a preact-bug when closing the Modal, and I'm too lazy to report it.

src/components/intents/IntentHandler.jsx
```diff
@@ -0,0 +1,93 @@
+import React, { Children, Component } from "react";
```
y-lohse: This whole file comes from cozy-store, and while the internals are what they are, the API is pretty sweet.

y-lohse:
[(2) Ajouter la fonctionnalité de fiche de contact personnelle pour The Machinery](https://trello.com/c/s038RXF2/955-2-ajouter-la-fonctionnalit%C3%A9-de-fiche-de-contact-personnelle-pour-the-machinery)

src/components/ContactsApp.jsx
```diff
@@ -24,7 +24,17 @@ const ContactsHeaderWithActions = ({ displayContactForm }, { t }) => (
...
+              label={"Select a Contact (intent)"}
+              action="PICK"
+            />
+          )}
+          {fakeintent !== null && (
+            <ContactsIntentButton
+              label={"Create a Contact (intent)"}
+              action="CREATE"
+              data={{ me: true }}
+            />
```
enguerran: To easily remove these lines later on, we should create a `<FakeIntentTollbar />` whose display is controlled by `fakeintent`.

src/components/ContactCard/ContactForm.jsx
```diff
@@ -77,6 +78,21 @@ const initialFieldValues = fields.reduce((initialValues, { name, isArray }) => {
   return initialValues;
 }, {});
 
+const ContactFormFooterWrapper = ({ portalInto, children }) => {
+  return portalInto ? (
+    <Portal into={portalInto}>{children}</Portal>
```
enguerran: Not sure why you use a portal, but, hey, if it works, that sounds good.

src/components/ContactsList/ContactRow.jsx
```diff
@@ -23,13 +28,19 @@ const ContactName = ({ firstname, lastname }) => (
...
+  me: PropTypes.bool
 };
 ContactName.defaultProps = {
   firstname: "",
-  lastname: ""
+  lastname: "",
+  me: false
 };
 
+const MeMarker = (props, { t }) => (
```
enguerran: Not really fan of this name, but it is hard to find a good name, maybe a `myselfTag` or `myselfLabel`.

src/components/ContactsList/ContactRow.jsx
```diff
@@ -88,7 +100,7 @@ const ContactRow = props => {
           onSelect={props.selection.onSelect}
         />
       )}
-      <ContactIdentity name={name} />
+      <ContactIdentity name={name} me={me} />
```
enguerran: Maybe _myself_ is better than _me_ to speak about _user identity_.

enguerran:
Ahah, MacOS filesystem is not case sensitive (https://github.com/cozy/cozy-contacts/pull/155/commits/d449b494b9b2d2306cefbd5eb33d3d71368ab7db) ;)

src/components/Intents/IntentHandler.jsx
```diff
@@ -0,0 +1,93 @@
+import React, { Children, Component } from "react";
```
y-lohse: Looks like my previous comments disappeared — this file is largely borrowed from cozy-store. It allows for a nice intent handling in the app.

src/components/ContactsApp.jsx
enguerran: tips: you can return a Array of components that will be added in the render tree as sibling.

```js
class ArrayOfHello extends React.Component {
  render() {
    return [...Array(100)].map((item, index) => (
      <Hello name={`Mister ${index}`} />
    ));
  }
}

const App = () => (
  <div style={styles}>
    <Hello name="CodeSandbox" />
    <h2>Start editing to see some magic happen {"\u2728"}</h2>
    {true && <ArrayOfHello />}
  </div>
);
```

https://codesandbox.io/s/q73kkr1k26

y-lohse: I don't think that works in preact. IIRc it's a react 16 feature that hasn't been ported yet!

enguerran: You are right I overzealously replaced the function with a functional component.
Here is the code with a function in preact:
https://codesandbox.io/s/v60w5ppm45

PR #220, opened 2018-08-03
chore: Optimize svg files with svgomg

enguerran:
see https://jakearchibald.github.io/svgomg/


src/assets/icons/calendar.svg
```diff
@@ -1,5 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
-  <g fill-rule="evenodd" transform="translate(-224 -192)">
-    <path fill="#95999D" d="M237,193 L240,193 L240,195 L224,195 L224,193 L227,193 L227,192 L229,192 L229,193 L235,193 L235,192 L237,192 L237,193 Z M224,196 L240,196 L240,206.997527 C240,207.551177 239.555369,208 238.999128,208 L225.000872,208 C224.448106,208 224,207.544239 224,206.997527 L224,196 Z M226,198 L226,206 L238,206 L238,198 L226,198 Z M232,201 L237,201 L237,205 L232,205 L232,201 Z"/>
-  </g>
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16">
```
GoOz: Removing `viewBox` attribute is not a good idea. Giving the circumstances, the SVG will not respect its own ratio. We had the issue lately. It may be too restrictive. SVGO doesn't remove it by default usually.

enguerran: I don't know enough of SVG rules to configure the svgomg tool. (for instance it says that `height` and `width` is redondant with `viewBox`. So I will close this PR as I cannot manage the good way to enhance SVG files.

GoOz:
Why did you close it? 

By definition `width`/`height` attributes take care of the viewport size, when `viewBox` tells what part of the viewport you should display. So by definition this it not redundant.

The thing is, the way we use SVGs for our icons, most of the time we want all of the viewport visible, hence the redundancy. 

If you really want the less possible code in your SVG, then there's a parameter in SVGOMG called "Prefer viewBox to width/height". That should do it.

If you want more info on that subject Sara Soueidan's posts on [24ways](https://24ways.org/2014/an-overview-of-svg-sprite-creation-techniques/) and, especially this one on coordinate system, on her [blog](http://www.sarasoueidan.com/blog/svg-coordinate-systems/) are really neat.

PR #226, opened 2018-08-09
chore: Use cozy-app-publish to publish app

y-lohse:


enguerran:
Hard to understand the purpose of this PR as there is no description
:trollface:

PR #247, opened 2018-08-29
refactor: Renamed variables

ptbrowne:
Renamed variables

- categorizedContacts -> contactsByLetter
- header -> letter

### DIFF
--- a/src/components/ContactsList/ContactsList.jsx
+++ b/src/components/ContactsList/ContactsList.jsx
@@ -15,7 +15,7 @@ const ContactsList = props => {
   }
   const sortedContacts = [...props.contacts].sort(sortLastNameFirst)
   let lastLetter = null
-  const categorizedContacts = sortedContacts.reduce((acc, contact) => {
+  const contactsByLetter = sortedContacts.reduce((acc, contact) => {
     const name = buildLastNameFirst(contact)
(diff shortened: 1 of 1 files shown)


src/components/ContactsList/ContactsList.jsx
```diff
@@ -28,11 +28,11 @@ const ContactsList = props => {
   return (
     <div className="list-wrapper">
       <ol className="list-contact">
-        {Object.keys(categorizedContacts).map(header => (
-          <li key={`cat-${header}`}>
-            <ContactHeaderRow key={header} header={header} />
+        {Object.keys(contactsByLetter).map(letter => (
```
enguerran: Thanks for the works.

The new proposed variable name describes what is stored in the variable and so it is strongly coupled with the implementation.

If the implementation of the old `categorizedContacts` changes, let say it is indexed by localization:

```js
{ paris:
    [ contact1
    , contact2
    , contact3
    ]
, toulouse:
    [ contact4
    , contact5
    , contact6
    ]
, marseille:
    [ contact7
    , contact8
    ]
}
```

The React Element hierarchy returned by the render function will still work, but the variable will not explain what really happen.

ptbrowne: OK. Fine by me.

PR #251, opened 2018-09-03
fix: Prevent toolbar from shrinking

y-lohse:
On small-ish screens, the toolbar would shrink and the list goes on top of the toolbar. This prevents the bar from shrinking below it's natural height.

### DIFF
--- a/src/styles/contacts.styl
+++ b/src/styles/contacts.styl
@@ -17,6 +17,7 @@
     max-width 80rem
     margin .75rem auto
     display flex
+    flex 0 0 auto
 
     .topbar__left
         flex 1


src/styles/contacts.styl
```diff
@@ -17,6 +17,7 @@
     max-width 80rem
     margin .75rem auto
     display flex
+    flex-shrink 0
```
enguerran: @GoOz I wonder what is recommended for flex rules:
- one rule `flex:` for all flex rules?
- separated `flex-` rules for readability?

GoOz: Well, it depends. 

I don't have any recommendation per se on that matter, more a personal preference but…

I tend to use `flex` shorthand when I want to set two or more properties or when I really want to lock things up and avoid some unexpected behaviour because [sadly every browser don't have the same default values for those three properties](https://github.com/philipwalton/flexbugs#flexbug-6) even if there is an official defaut: `flex: 0 1 auto`. (I know only older are concerned but a little more explicit code for safety doesn't kill)

And I use `flex-` properties mostly for overwriting.

In that case, because the element doesn't have `flex-basis` and `flex-grow` set, I would have written it like this `flex: 0 0 auto` for safety.

PR #253, opened 2018-09-03
fix: Reset input to fire onChange on Chrome

enguerran:
Changes:

10a3287 (Enguerran, 27 seconds ago)
   refactor: Rename a variable

   There is two variables with the same name but different cases. It is not
   very easy to read.

a3d5b7a (Enguerran, 2 minutes ago)
   fix: Reset input to fire onChange on Chrome

   On Chrome, `onChange` is not fired when the selected file is the same (I
   guess there is no real change for webkit). See
   https://stackoverflow.com/questions/6623310/input-file-onchange-event-not-being-fired-in-chrome
   Using a controlled component to handle input state is a good way in React
   to reset the input value when needed.

   fix 252


src/components/ContactImportationModal/ContactImportationCard/ImportationInput.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+class ImportationInput extends React.Component {
+  state = {
+    value: ''
+  }
+
+  handleOnChange = event => {
+    const file = event.target.files[0]
+    this.props.fileAction(file)
+    this.setState(state => ({ ...state, value: '' }))
```
ptbrowne: OK understood. Nice !

Since it is a bit surprising, I think it would be good to have the link to the stackoverflow answer you pointed in the chat :)

enguerran: I put the link into the commit's message.
I'm still not a big fan of comments in the code because they are not historized, while git history (with a `git blame` in the way "but who did this?") lets you know what happened.

RT if you agree
fav if you still want me to add a comment
(or just reply below because we are not on twitter)

ptbrowne: > I'm still not a big fan of comments in the code because they are not historized,

In what sense are they not historized ? They are historized in the SCM right ?

I understand that we should not have our code littered with stackoverflow answers but I think this one is particularly surprising and would qualify for it. I am not sure I would have `git blame`d this code right away. I think I would have had trouble to understand first.

enguerran: I added a comment

PR #258, opened 2018-09-06
feat: Update vcard merge strategies

enguerran:
I added some tests to cover the merge strategies.
I don't know if we may use jest Snapshot or JS expected object.
As I wrote the merge code with TDD, I did not use Jest Snapshot.

### DIFF
--- a/src/connections/allContacts.js
+++ b/src/connections/allContacts.js
@@ -1,11 +1,16 @@
 import { connect, withMutations } from 'cozy-client'
 import flow from 'lodash/flow'
-import array from 'lodash/array'
+import uniqWith from 'lodash/uniqWith'
 import isEqual from 'lodash/isEqual'
 import flatten from 'lodash/flatten'
+import mergeWith from 'lodash/mergeWith'
+import merge from 'lodash/merge'
+import concat from 'lodash/concat'
+import groupBy from 'lodash/groupBy'
+import map from 'lodash/map'
 
 function removeDuplicates(data) {
-  return array.uniqWith(data, isEqual)
+  return uniqWith(data, isEqual)
 }
 
 function findItems(doctype, selectors) {
@@ -62,11 +67,7 @@ export const withContactsMutations = withMutations(client => ({
   importContact: async attributes => {
     const contacts = await findContactsWithSamePhoneOrEmail(attributes)(client)
     if (contacts.length === 1) {
-      // TODO: create a more complex updated contact if we need to merge deep properties together.
-      const updatedContact = {
-        ...contacts[0],
-        ...attributes
-      }
+      const updatedContact = mergeContact(contacts[0], attributes)
       return client.save(updatedContact)
     } else if (contacts.length > 1) {
       throw new Error(
@@ -104,3 +105,45 @@ export const withContactsMutations = withMutations(client => ({
 }))
 
 export default flow([withContacts, withContactsMutations])
+
+export function mergeContact(contact, attributes) {
+  // TODO: refactor to remove customizer from the mergeContact
+  const propertyName = {
+    email: 'address',
+    phone: 'number'
+  }
+
+  function customizer(objValue, srcValue, key) {
+    if (Array.isArray(objValue) || Array.isArray(srcValue)) {
+      return mergeArrayOnKey(objValue, srcValue, propertyName[key])
+    }
+    if (key === 'fullname' && objValue && srcValue) {
+      return objValue.length > srcValue.length ? objValue : srcValue
+    }
+  }
+  return mergeWith({}, contact, attributes, customizer)
+}
+
+/**
+ *
+ * @param {Array} objValue
+ * @param {Array} srcValue
+ * @param {String} key
+ *
+ * var arr1 = [{ x: 1 }];
+ * var arr2 = [{ x: 2 }];
+ * mergeArrayOnKey(arr1, arr2, 'x')
+ * → [{ x: 1 }, { x: 2 }]
+ *
+ * or
+ *
+ * var arr1 = [{ x: 1, y: 1 }];
+ * var arr2 = [{ x: 1, z: 1 }, { x: 2, z: 2 }];
+ * mergeArrayOnKey(arr1, arr2, 'x')
+ * → [{ x: 1, y: 1, z: 1 }, { x: 2, z: 2 }]
+ */
+function mergeArrayOnKey(objValue = [], srcValue = [], key) {
+  return map(groupBy(concat(objValue, srcValue), key), vals =>
+    merge({}, ...vals)
+  )
+}
--- a/src/connections/allContacts.js
+++ b/src/connections/allContacts.js
@@ -1,58 +1,7 @@
 import { connect, withMutations } from 'cozy-client'
 import flow from 'lodash/flow'
-import uniqWith from 'lodash/uniqWith'
-import isEqual from 'lodash/isEqual'
-import flatten from 'lodash/flatten'
-import mergeWith from 'lodash/mergeWith'
-import merge from 'lodash/merge'
-import concat from 'lodash/concat'
-import groupBy from 'lodash/groupBy'
-import map from 'lodash/map'
-
-function removeDuplicates(data) {
-  return uniqWith(data, isEqual)
-}
-
-function findItems(doctype, selectors) {
-  return client =>
-    Promise.all(
-      selectors.map(selector =>
-        client.query(client.find(doctype).where(selector))
-      )
-    )
-      .then(results => results.map(result => result.data))
-      .then(flatten)
-      .then(removeDuplicates)
-}
-
-function selector(attr, prop) {
-  return values => ({
-    [attr]: {
-      $elemMatch: {
-        [prop]: {
-          $in: values
-        }
-      }
-    }
-  })
-}
-
-function getValues(array = [], property) {
-  return array.map(item => item[property])
-}
-
-function createSelector(object, path) {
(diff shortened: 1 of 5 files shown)


src/connections/allContacts.js
```diff
@@ -104,3 +102,18 @@ export const withContactsMutations = withMutations(client => ({
 }))
 
 export default flow([withContacts, withContactsMutations])
+
+export function mergeContact(contact, attributes) {
+  function concat(arr1 = [], arr2 = []) {
+    return [...arr1, ...arr2]
+  }
+  function customizer(objValue, srcValue, key) {
```
sebn: What about making it top-level, so we could unit-test attribute merging logic on its own without having to deal with a full contact object in every test?

Although some people consider that *private* functions should not be exposed for testing purpose, maybe it's worth the trouble.

Array attributes would probably still need to be tested as a whole in order to be able to either merge two corresponding items or keep them separate depending on their content.

src/connections/allContacts.js
```diff
@@ -104,3 +102,18 @@ export const withContactsMutations = withMutations(client => ({
...
+    if (Array.isArray(objValue) || Array.isArray(srcValue)) {
+      return concat(objValue, srcValue)
```
sebn: We'll probably want to prevent duplicates at some point, but I assume this will happen in another PR :slightly_smiling_face:

src/connections/allContacts.js
```diff
@@ -104,3 +102,18 @@ export const withContactsMutations = withMutations(client => ({
...
+    }
+    if (key === 'fullname') {
+      return objValue.length > srcValue.length ? objValue : srcValue
```
sebn: Wouldn't this make data loss possible?
E.g. `{fullname: "John J. Doe", ...}` and `{fullname: "John Doe Jr", ...}`? (sorry, I can't think of a better example)

What about checking whether some of the strings include the other one, defaulting to both slash-concatenated? It should work assuming the attributes containers (contacts here) are considered identicals (i.e. with the same identity, not necessarily the same value).

src/connections/allContacts.js
```diff
@@ -104,3 +102,18 @@ export const withContactsMutations = withMutations(client => ({
...
+    }
+  }
+  return mergeWith({ ...contact }, attributes, customizer)
```
sebn: :+1: `mergeWith` looks like a good solution!

src/connections/allContacts.js
```diff
@@ -104,3 +102,18 @@ export const withContactsMutations = withMutations(client => ({
 }))
 
 export default flow([withContacts, withContactsMutations])
+
+export function mergeContact(contact, attributes) {
+  function concat(arr1 = [], arr2 = []) {
+    return [...arr1, ...arr2]
```
sebn: :ok_hand: It's weird that neither `Array#concat()`, `Array.from()`, `_.concat()` nor the spread operator handle a `null` value as an empty sequence :confused:

src/connections/allContacts.js
```diff
@@ -104,3 +102,18 @@ export const withContactsMutations = withMutations(client => ({
...
+  }
+  function customizer(objValue, srcValue, key) {
+    if (Array.isArray(objValue) || Array.isArray(srcValue)) {
+      return concat(objValue, srcValue)
+    }
+    if (key === 'fullname') {
```
sebn: Couldn't this be generalized to other string attributes? (I'm not sure, just asking)

Also what happens to them with the current implementation?

sebn:
> I don't know if we may use jest Snapshot or JS expected object.
> As I wrote the merge code with TDD, I did not use Jest Snapshot.

In my initial implementation, I used both:
- Snapshots in integration tests to have full output examples and prevent regressions (but I finally excluded them from the PR according to feedback)
- Unit tests for (almost) every function, testing edge cases at the lowest possible level

Should I have to do it again, I believe I'd go the same way.

src/connections/allContacts.js
sebn: (this one sound like a blocker)

src/connections/allContacts.js
sebn: Actually I don't remember of other top-level string attributes... Are there any?

sebn:
Also maybe we could introduce some merge idempotence testing (i.e. making sure that merging the same contacts twice ends up with the same result)?

src/connections/allContacts.js
enguerran: Ohhh, you are so right.
No, I want to add that to this PR.

src/connections/allContacts.js
enguerran: Ahhh EcmaScript :disappointed:
:rofl:

src/connections/allContacts.js
enguerran: I understand your proposal, however:
The added value of the functionality is the merge of contacts, a unit test on an implementation detail does not seem easy to maintain.

enguerran: @sebn I refactored the way I manage merge strategies. Can you take a look at it?

src/connections/allContacts.js
enguerran: In https://github.com/cozy/cozy-doctypes/pull/57 I suggest to remove this field.

And I changed the target of `takeLonger` strategy to become `givenName` and `familyName`.

enguerran:
@sebn I refactored the way I manage merge strategies. Can you take a look at it?

src/connections/allContacts.js
```diff
@@ -104,3 +105,45 @@ export const withContactsMutations = withMutations(client => ({
...
+export function mergeContact(contact, attributes) {
+  // TODO: refactor to remove customizer from the mergeContact
+  const propertyName = {
+    email: 'address',
+    phone: 'number'
+  }
+
+  function customizer(objValue, srcValue, key) {
+    if (Array.isArray(objValue) || Array.isArray(srcValue)) {
+      return mergeArrayOnKey(objValue, srcValue, propertyName[key])
```
y-lohse: What happens if `propertyName[key]` is `undefined`? Can that happen?

enguerran: I changed the implementation, can you check if it is all good.
If there is no strategie, the classical lodash's merge is used.

enguerran:
@y-lohse @sebn can you check if the new implementation looks good to you. Regarding the tests, they cover the cases in mind when implementing the features. If we want some additional check, we juste have to add a test and implement in TDD mode.

src/helpers/mergeContact.js
```diff
@@ -0,0 +1,55 @@
...
+  email: mergeArrayOfObject('address'),
+  phone: mergeArrayOfObject('number'),
+  givenName: keepLonger,
+  familyName: keepLonger
+}
+
+const fromStrategies = strategies => (objValue, srcValue, key) =>
+  strategies[key] ? strategies[key](objValue, srcValue, key) : undefined
+
+const customizer = fromStrategies(strategies)
```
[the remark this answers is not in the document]
enguerran: It is written for readability and understandability. The named function that takes an object `strategies` and returns a function with 3 parameters `objValue, srcValue, key` and returns `undefined` or a property of the object `strategies` express its intention with this naming.

src/helpers/mergeContact.js
```diff
@@ -0,0 +1,55 @@
...
+
+function mergeArrayOfObject(key) {
+  return (objValue, srcValue) => {
+    return mergeArrayOnKey(objValue, srcValue, key)
+  }
+}
+
+function keepLonger(objValue = '', srcValue = '') {
+  return objValue.length > srcValue.length ? objValue : srcValue
```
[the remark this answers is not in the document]
enguerran: We use the length of the string as a basis for the strategy that wants to keep the longest string.

The Product Owner wrote a story that says the user wants to keep the longest chain because it is a good heuristic to keep the richest information.
Unless unrealistic examples such as `"John''` vs `"AAAAAAAAAAAAA''`, we can consider that `"John''` contains less information than `"Lord Commander John''` or `"Jonathan''`.

src/helpers/mergeContact.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    const fakeContact = {
+      email: [{ address: 'starck@localhost' }],
+      name: { familyName: 'Starck' },
+      phone: [{ number: '0987654321', primary: true, type: 'Work' }],
+      address: [{ formattedAddress: 'Westeros', primary: true, type: 'Work' }]
+    }
+    const importedContact = {
+      name: { familyName: 'Starck', givenName: 'Ned' },
+      phone: [{ number: '0123456789' }],
+      email: [{ address: 'ned@localhost' }]
```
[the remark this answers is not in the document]
enguerran: In this test, we are given two contacts and check the merge with the expected merged contact. We don't decide if the contact should be merged or not. But you are right, we could add a property common to both contacts, but it will not change the test expectation.

src/connections/allContacts.js
```diff
@@ -62,11 +16,7 @@ export const withContactsMutations = withMutations(client => ({
...
     if (contacts.length === 1) {
-      // TODO: create a more complex updated contact if we need to merge deep properties together.
-      const updatedContact = {
-        ...contacts[0],
-        ...attributes
-      }
+      const updatedContact = mergeContact(contacts[0], attributes)
       return client.save(updatedContact)
     } else if (contacts.length > 1) {
       throw new Error(
```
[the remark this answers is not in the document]
enguerran: I guess someday the application will have a merging tool.

PR #276, opened 2018-09-25
chore: prepare release 0.6.1

enguerran:
- fix: Reset input to fire onChange on Chromium
- fix: Prevent toolbar from shrinking
- fix: Error when creating a new contact if list is empty
- feat: Update vcard merge strategies
- feat: Remove fullname from schema


Crash--:
👍

PR #279, opened 2018-10-01
Use app-name-prefix in .ejs template

CPatchane:
Used by the bar to display 'Cozy Contacts'


nono:
Can we merge this PR?

PR #289, opened 2018-10-23
fix: Send cookies when fetching manifest.json

nono:
See https://www.w3.org/TR/appmanifest/#obtaining


nono:
@enguerran yes, the manifest is not in a public directory, so we need to send the cookies to avoid a 401 error.

PR #293, opened 2018-10-25
[WIP] Chore/upgrade client

Crash--:


src/components/ContactsList/ContactHeaderRow.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const ContactHeaderRow = props => <div className="divider">{props.header}</div>
+
+ContactHeaderRow.propTypes = {
+  header: PropTypes.string.isRequired
+}
+export default ContactHeaderRow
```
y-lohse: 👍

src/components/ContactsList/ContactRow.jsx
```diff
@@ -68,7 +68,7 @@ const ContactSelection = props => (
     }}
   >
     <span data-input="checkbox">
-      <input type="checkbox" checked={props.selected} />
+      <input type="checkbox" checked={props.selected} readOnly />
```
y-lohse: Why do we need the `readonly` attribute now?

src/components/ContactsList/ContactsList.jsx
```diff
@@ -1,66 +1,117 @@
-import React from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { sortLastNameFirst, buildLastNameFirst } from './'
 import ContactsEmptyList from './ContactsEmptyList'
 import ContactRow from './ContactRow'
+import ContactHeaderRow from './ContactHeaderRow'
+import { Query } from 'cozy-client'
+import ContactCardModal from '../Modals/ContactCardModal'
+import _ from 'lodash'
```
y-lohse: It's preferable to import only the functions we need from lodash, rather than the whole package. So something like:

```suggestion
import find from 'lodash/find'
```

src/components/ContactsList/ContactsList.jsx
```diff
@@ -1,66 +1,117 @@
...
+                      )}
+                      groups={groups}
+                    />
+                  )}
+                </div>
+              </div>
+            )
+          }
+        }}
+      </Query>
```
y-lohse: Not blocking for this PR, but I think we should split this component next chance we get — separating the loading state, the modal management, etc.

src/components/Modals/ContactCardModal.jsx
```diff
@@ -16,6 +16,8 @@ const ContactCardMenu = ({ deleteAction }) => (
         extension="narrow"
         icon="dots"
         className="fix-c-btn"
+        iconOnly
+        label={'deleteAction.label'}
```
y-lohse: We're missing a call to `t` here :)

src/store/configureStore.js
```diff
@@ -0,0 +1,46 @@
+/* global __DEVELOPMENT__ */
+import { compose, createStore, applyMiddleware } from 'redux'
+import thunkMiddleware from 'redux-thunk'
+import { createLogger } from 'redux-logger'
+import flag from 'cozy-flags'
```
y-lohse: Are we using this?

src/targets/browser/index.jsx
```diff
@@ -7,15 +7,21 @@ import CozyClient, { CozyProvider } from 'cozy-client'
...
 
 const RootApp = props => (
-  <I18n lang={props.lang} dictRequire={lang => require(`locales/${lang}`)}>
-    <CozyProvider client={props.client}>
-      <App />
-    </CozyProvider>
-  </I18n>
+  <Provider store={props.store}>
+    <I18n lang={props.lang} dictRequire={lang => require(`locales/${lang}`)}>
+      <CozyProvider client={props.client}>
```
y-lohse: The `CozyProvider` can also be used to provide the store, which also avoids having 2 separate stores for cozy-client and the rest of the app. You can check out how it's done in drive :)

src/targets/browser/index.jsx
```diff
@@ -7,15 +7,21 @@ import CozyClient, { CozyProvider } from 'cozy-client'
...
+        <HotedApp />
+      </CozyProvider>
+    </I18n>
+  </Provider>
 )
 
+const HotedApp = hot(module)(App)
```
y-lohse: Does this have any side effects in production?

src/components/ContactCard/ContactGroups.jsx
```diff
@@ -2,27 +2,33 @@ import React from 'react'
...
 import { fullContactPropTypes } from '../ContactPropTypes'
 import ContactGroupManager from '../ContactGroups/ContactGroupManager'
-import { withGroups } from '../../connections/allGroups'
 import { withContactsMutations } from '../../connections/allContacts'
-import get from 'lodash/get'
-
-export class ContactGroups extends React.Component {
+import { Query } from 'cozy-client'
+const groupsQuery = client => client.all('io.cozy.contacts.groups')
+class ContactGroupsClass extends React.Component {
```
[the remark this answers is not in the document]
Crash--: I was lazy. I didn't want to rewrite all the imports ^-^'

src/components/ContactCard/ContactGroups.jsx
```diff
@@ -2,27 +2,33 @@ import React from 'react'
...
   updateContactGroups = groups => {
-    const modifiedContact = { ...this.props.contact }
-    modifiedContact.relationships.groups.data = groups
-    this.props.updateContact(modifiedContact)
+    const { contact } = this.props
+    if (!contact.groups) {
+      contact.groups = {}
+    }
+    contact.groups.data.map(group => {
+      contact.groups.removeById(group._id)
```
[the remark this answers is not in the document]
Crash--: good catch! Thanks

src/store/configureStore.js
Crash--: Not yet! But it should be added in our skeleton :D

package.json
```diff
@@ -49,16 +50,20 @@
...
-    "preact-compat": "3.18.0",
+    "react": "^16.5.2",
     "react-final-form": "3.2.1",
-    "react-final-form-arrays": "1.0.6"
+    "react-final-form-arrays": "1.0.6",
+    "react-hot-loader": "^4.3.11",
+    "react-redux": "^5.0.7",
+    "redux": "^4.0.1",
+    "redux-logger": "^3.0.6",
+    "redux-thunk": "^2.3.0"
```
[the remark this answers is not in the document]
y-lohse: Sorry, I saw that later and forgot to delete the comment^^

src/components/Buttons/ImportGoogleButton.jsx
```diff
@@ -25,4 +28,4 @@ ImportGoogleButton.defaultProps = {
   onComplete: () => {}
 }
 
-export default ImportGoogleButton
+export default translate()(ImportGoogleButton)
```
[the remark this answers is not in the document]
y-lohse: Required with the switch to react, at least for now.

src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -19,14 +19,6 @@ class ContactFieldInput extends React.Component {
...
   }
 
-  onMainInputBlur = e => {
-    this.setState(state => ({
-      ...state,
-      isRenderingLabel: e.target.value && this.props.withLabel,
-      hasFocus: false
-    }))
-  }
-
```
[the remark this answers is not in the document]
y-lohse: Causes a new bug, will be corrected later.

src/components/ContactsList/ContactRow.jsx
y-lohse: Required by cozy-ui if we don't have an event handler.

src/targets/browser/index.jsx
y-lohse: No.

PR #298, opened 2018-10-29
refactor: Remove ...state and use function call on setState

Crash--:


enguerran:
I cannot understand the purpose of these commit and PR.
The type of the commit is `chore`, but
> chore: updating build tasks, package manager configs, etc; no production code change

See https://github.com/cozy/cozy-guidelines#type

The subject of the commit is not _"Remove ...state"_, it is: _"Use a different setState that can lead to asynchronous error"_.

See React documentation:
> we recommend using the updater function form, instead:

More details in this article: https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b

The code you update is not using the previous state value, but for homogeneity and idiomatic codebase, it appears to be a good practice to use the function form of `setState`. What do you think?

Crash--:
- commit message ✅ 
- setState callback ✅  (Next time I'll let you update your own files with that change ;) #vengeance )

ptbrowne:
If the setState is not using the previous state, I personally see no problem using the object form of setState. I find it simpler to use and read.

The sentence in React documentation is "If the next state depends on the current state, we recommend using the updater function form".

One advantage stated in the article in favor of the functional `setState` is the ability to extract the updater from the component classes. This seems interesting as it can lead to better testability. But it comes with decreased readibility since you have to look at two places to understand what's going on. Since the setState are very simple, I see no big advantage for testability vs readibility here, when the setState are very simple.

PR #299, opened 2018-10-30
fix: Vcard import message

Crash--:


Crash--:
[vCard - message confirmation - regression](https://trello.com/c/tcNGENtR/1368-vcard-message-confirmation-regression)

Crash--:
Maj Transifex FR ✅

src/components/ContactImportationModal/index.jsx
```diff
@@ -47,15 +47,19 @@ class ContactImportationModal extends React.Component {
...
       const created = Importation.created(finishedImportation)
       const updated = Importation.updated(finishedImportation)
+      const stringCreated =
+        created > 0 ? t('importation.creation', created) : ''
+      const stringUpdated =
+        updated > 0
+          ? created > 0
+            ? `, ${t('importation.update', updated)}`
+            : `${t('importation.update', updated)}`
+          : ''
```
y-lohse: Not super fond of double ternaries, but I'm having a hard time coming up with a good alternative.

PR #300, opened 2018-10-30
feat: Add Spinner Component

Crash--:


Crash--:
[Contacts - chargement de la liste](https://trello.com/c/Yg57laB0/1280-contacts-chargement-de-la-liste)

PR #306, opened 2018-10-31
fix: No need for : but !

Crash--:


src/locales/en.json
```diff
@@ -29,7 +29,7 @@
   "importation": {
     "cancel": "Cancel",
     "complete_failure": "Import failed.",
-    "complete_success": "Congratulations",
+    "complete_success": "Congratulations!",
     "creation": "a contact has been created |||| %{smart_count} contacts have been created",
```
enguerran: This message may start with an uppercase.

src/locales/en.json
```diff
@@ -29,7 +29,7 @@
...
     "update": "a contact has been updated |||| %{smart_count} contacts have been updated",
```
enguerran: itou

PR #310, opened 2018-11-02
refacto: Use ModalManager

Crash--:
In this PR I use the `showModal` to display the 2 modals (import / create). This is the responsibility of the Toolbar to deal with that, not the `App` component 

On large list, this is a big perf improvement since opening these 2 modals doesn't refresh the all list.

I had to comment the blur _again_. Will find out why.

- [x] In a second time, I'll need to address the issue of opening the Modal to display a contact. 

- [x] In a third time, I'll fix the selection action. ==> No need to render everything ;) => Done. I think we can do better, even after my first attempt we still re-render all the checkbox. (but only the checkboxes )

- [x] restore onBlur 

- [x] fix tests

Next possible actions : 
- No need to refresh all the select bar when toggling selection 
- ATM when we change the group of a contact, there is a "global" refresh
- ContactForm needs to be rewritten


src/components/Toolbar.jsx
```diff
@@ -1,33 +1,71 @@
...
 import { PropTypes } from 'prop-types'
 import { Button, Icon } from 'cozy-ui/react'
-import IconTeam from '../assets/icons/team.svg'
+
 import { translate } from 'cozy-ui/react/I18n'
+import { Query } from 'cozy-client'
+
+import ContactFormModal from './Modals/ContactFormModal'
+import withModal from './HOCs/withModal'
+import ContactCardModalConnected from './Modals/ContactCardModalConnected'
```
y-lohse: the file `ContactCardModalConnected ` has been deleted in this PR..?

Crash--: Yep! But you're commenting an 'Outdated' code. :D

src/components/Selection/selectionContainer.js
```diff
@@ -0,0 +1,30 @@
+import { connect } from 'react-redux'
```
y-lohse: I think this file can simply be called `container`, since it's inside the `selection` folder? And same thing for the `selectionReducer`.

PR #315, opened 2018-11-07
fix: Don't crash if contact doesn't have some info

Crash--:
If contact didn't have a `name` then the app crashed.


Crash--:
[(1) Contacts: Utilisation impossible car ça rame trop](https://trello.com/c/ne1TlRak/1376-1-contacts-utilisation-impossible-car-%C3%A7a-rame-trop)

Crash--:
added a test

y-lohse:
Fantastic, thanks!

PR #316, opened 2018-11-07
Feat/create group

Crash--:


src/components/ContactGroups/ContactGroupCreation.jsx
```diff
@@ -0,0 +1,96 @@
...
+      this.textInput.current.value = ''
+    }
+    if (e.keyCode === 8) {
+      e.stopPropagation()
+    }
+    //space key is used to validate an option
+    if (e.keyCode === 32) {
+      e.stopPropagation()
+    }
+    e.stopPropagation()
```
y-lohse: I think we can remove the previous calls to `stopPropagation` since this one is always called?

src/components/ContactGroups/ContactGroupCreation.jsx
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    name: name,
+    metadata: {
+      version: 1
+    }
+  }
+}
+class ContactGroupCreation extends Component {
+  state = {
+    isInput: false,
```
y-lohse: Can we find a better name for `isInput`? `isInputActive` or `isInputDisplayed` maybe?

src/components/ContactGroups/ContactGroupManager.jsx
```diff
@@ -1,46 +1,63 @@
...
-  </Button>
-)
+const MainButton = ({ t, selectProps: { toggleMenuIsOpen } }) => {
+  return (
+    <Button
+      theme="secondary"
+      size="small"
+      icon={<Icon icon={IconGroups} color={palette['coolGrey']} />}
+      label={t('groups.manage')}
+      tabIndex={1}
```
y-lohse: Do we still want this `tabIndex`?

src/components/ContactGroups/ContactGroupManager.jsx
```diff
@@ -51,33 +68,75 @@ const captureEscapeEvent = e => {
...
+      createGroup,
+      t
+    } = this.props
+
+    return (
+      <>
+        {this.state.menuIsOpen && (
+          <Overlay
+            className={classNames('overlay-creation-group')}
+            onClick={() => this.toggleMenuIsOpen()}
```
y-lohse:
```suggestion
            onClick={this.toggleMenuIsOpen}
```

src/components/ContactGroups/ContactGroupManager.jsx
```diff
@@ -1,46 +1,63 @@
...
+  }),
+  menu: base => ({
+    ...base
+  }),
+  noOptionsMessage: () => ({}),
+  option: base => ({
+    ...base,
+    whiteSpace: 'nowrap',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis'
```
y-lohse: Does this work in the end?

src/components/ContactGroups/index.styl
```diff
@@ -0,0 +1,34 @@
+@require 'cozy-ui'
+@require 'settings/palette.styl'
+
+.container
+  height rem(48)
+  display: flex;
```
y-lohse:
```suggestion
  display flex
```

src/components/ContactGroups/index.styl
```diff
@@ -0,0 +1,34 @@
...
+  align-items: center;
```
y-lohse:
```suggestion
  align-items center
```

src/components/ContactCard/ContactGroups.jsx
```diff
@@ -20,18 +24,44 @@ class ContactGroupsClass extends React.Component {
     })
     this.props.updateContact(contact)
   }
+  componentDidUpdate(prevProps, prevState) {
+    if (prevState.createdGroup.length !== this.state.createdGroup.length) {
+      document.getElementsByClassName('react-select__menu-list')[0].scrollTop =
```
y-lohse: Do you think we could make this work in the `ContactGroupManager`? As it is, we're creating coupling between these 2 components via a class name, which seems fragile.
When a group is added, the `ContactGroupManager ` receives a props update so in theory it should be possible to do the scroll there?

src/components/ContactGroups/ContactGroupCreation.jsx
```diff
@@ -0,0 +1,96 @@
+import React, { Component } from 'react'
+import { Icon, Input, translate } from 'cozy-ui/react'
+import classNames from 'classnames'
+import './index.styl'
```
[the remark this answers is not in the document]
Crash--: I was doing some tests with `React.lazy` ;)

src/components/ContactGroups/ContactGroupManager.jsx
```diff
@@ -1,46 +1,62 @@
...
-const MainButtonControl = reactSelectControl(<MainButtonWithTranslation />)
-
+const MainButtonControl = props => <MainButtonWithTranslation {...props} />
 const customStyles = {
   container: base => ({
     ...base,
     display: 'inline-block',
-    verticalAlign: 'middle'
+    verticalAlign: 'middle',
+    zIndex: 100000
```
y-lohse: I suggest we make the following change:

```
const zIndexAboveOverlay = 51 // see https://github.com/cozy/cozy-ui/blob/master/stylus/settings/z-index.styl#L42
…
verticalAlign: 'middle',
    zIndex: zIndexAboveOverlay
```

So we know why we have a z-index here, and it also prevents going going over things we don't want.

src/components/ContactGroups/ContactGroupManager.jsx
```diff
@@ -1,46 +1,62 @@
...
+  }),
+  menu: base => ({
+    ...base
+  }),
+  noOptionsMessage: () => ({}),
```
y-lohse: Why do we need to reset the styles of this component?

Crash--: Because I wrote all the needed style here: https://github.com/cozy/cozy-contacts/pull/316/files#diff-0f0d1cddc45a5df3cf945d6f6311fab7R29 and didn't want to add !important everywhere ;)

PR #320, opened 2018-11-09
fix: Import when empty

Crash--:


Crash--:
[Contacts - liste vide bouton import](https://trello.com/c/8fDt68c8/1405-contacts-liste-vide-bouton-import)

PR #329, opened 2018-11-13
fix: PO feedbacks

Crash--:
- Change string for no groups 
- Don't allow the creation of a seconde group with an already used name 
- Set modal in fullscreen on mobile 
- Align text to initial. No need to center it :o

### DIFF
--- a/src/components/ContactCard/ContactIdentity.spec.jsx
+++ b/src/components/ContactCard/ContactIdentity.spec.jsx
@@ -26,7 +26,6 @@ describe('ContactIdentity', () => {
         <ContactIdentity contact={contactWithoutName} groups={groupsMock} />
       </AppLike>
     )
-    //console.log('test', app.debug())
     const avatar = app.find('Avatar')
     expect(avatar.prop('text')).toEqual('SF')
   })
@@ -52,7 +51,6 @@ describe('ContactIdentity', () => {
         <ContactIdentity contact={contactWithoutName} groups={groupsMock} />
       </AppLike>
     )
-    //console.log('test', app.debug())
     const avatar = app.find('Avatar')
     expect(avatar.prop('text')).toEqual('')
   })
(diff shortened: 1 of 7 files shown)


src/components/ContactCard/ContactIdentity.spec.jsx
```diff
@@ -26,7 +26,7 @@ describe('ContactIdentity', () => {
         <ContactIdentity contact={contactWithoutName} groups={groupsMock} />
       </AppLike>
     )
-    //console.log('test', app.debug())
+    //'test', app.debug())
```
y-lohse:
```suggestion
```

src/components/ContactGroups/helpers/groups.js
```diff
@@ -0,0 +1,12 @@
+export const checkIfGroupAlreadyExists = (
+  groupsAlreadyCreated,
+  groupToCreate
+) => {
+  let alreadyThisName = false
```
y-lohse: How about something like this:

```
groupsAlreadyCreated.find(group => group.name.toLowerCase() === groupToCreate.name.toLowerCase()) !== undefined
```

I think it works the same and is a bit easier to read? Otherwise I would suggest we rename `alreadyThisName` to something like `isNameAlreadyUsed `

src/components/ContactCard/ContactIdentity.spec.jsx
Crash--: OMG c'est quoi cette suppression de ligne qui n'a pas fonctionné :'(

src/components/ContactGroups/helpers/groups.js
```diff
@@ -0,0 +1,11 @@
+export const checkIfGroupAlreadyExists = (
+  groupsAlreadyCreated,
+  groupToCreate
+) => {
+  const isNameAlreadyUsed =
+    groupsAlreadyCreated.find(
+      group => group.name.toLowerCase() === groupToCreate.name.toLowerCase()
+    ) !== undefined
+
+  return isNameAlreadyUsed
```
y-lohse: I think we could have just returned the result of the `find`, but whatever :)

PR #330, opened 2018-11-13
fix: Open the right intent

Crash--:


Crash--:
[Google - modale blanche - regression Home](https://trello.com/c/Z7BfL4OQ/1411-google-modale-blanche-regression-home)

PR #331, opened 2018-11-13
chore: Use renovate cozy config

kosssi:


y-lohse:
Thanks! We still want to automerge devDependencies here, i'll amend the PR.

PR #352, opened 2018-11-29
chore: Explicitly use react in tests

y-lohse:
Our tests were running using preact instead of react. Usually not a big problem, but now that [`preact-portal`](https://github.com/cozy/cozy-ui/pull/723) is a peer dependency [it was causing errors](https://travis-ci.org/cozy/cozy-contacts/jobs/461227033#L2106).


Crash--:
Nice catch!

PR #354, opened 2018-11-30
Properly delete contact groups

y-lohse:
The groups where not deleted in the right field, which was causing problems.

There's still the underlying question of why groups seem to get saved in the `relationships` AND `groups` property, which I haven't investigated yet.

### DIFF
--- a/src/thunks/cleanTrashedGroups.js
+++ b/src/thunks/cleanTrashedGroups.js
@@ -16,30 +16,34 @@ const cleanTrashedGroups = () => async (dispatch, getState, { client }) => {
 }
 
 const removeGroupFromAllContacts = async (client, groupId) => {
-  const { data: contacts, next: hasMore } = await client.query(
-    client.find(DOCTYPE_CONTACTS).where({
-      groups: {
-        $elemMatch: {
-          _id: groupId,
-          _type: DOCTYPE_CONTACT_GROUPS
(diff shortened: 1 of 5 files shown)


src/thunks/cleanTrashedGroups.js
```diff
@@ -18,10 +18,14 @@ const cleanTrashedGroups = () => async (dispatch, getState, { client }) => {
...
-          _id: groupId,
-          _type: DOCTYPE_CONTACT_GROUPS
+      relationships: {
+        groups: {
+          data: {
+            $elemMatch: {
+              _id: groupId,
+              _type: DOCTYPE_CONTACT_GROUPS
+            }
+          }
```
Crash--: I think this is where and why you need the use of `.includes('groups')`, nop ?

y-lohse: No, I don't actually care about the groups themselves right there, I just need the ids. Either that or I misunderstood what ypu meant.

y-lohse:
@Crash-- I updated the PR — we're not manipulating the `relationships` object anymore, but we still have to query it manually.
I made the changes so we don't need to update cozy-client, but I'll create issues there to see what is best long term.

y-lohse:
[(3) Groupes - supprimer](https://trello.com/c/AUW3POQY/1404-3-groupes-supprimer)

PR #413, opened 2019-01-17
Feat/edit: contact

cedricmessiant:


cedricmessiant:
[(2) éditer un contact](https://trello.com/c/Zf79Mvzf/717-2-%C3%A9diter-un-contact)

src/components/ContactCard/ContactForm.jsx
```diff
@@ -77,7 +77,62 @@ const initialFieldValues = fields.reduce((initialValues, { name, isArray }) => {
   return initialValues
 }, {})
```
Crash--: file to rename ? I like having component name === file name :-°

src/components/ContactCard/ContactForm.jsx
```diff
@@ -77,7 +77,62 @@ const initialFieldValues = fields.reduce((initialValues, { name, isArray }) => {
...
 
-class ContactForm extends React.Component {
+export const moveToHead = shouldBeHead => items =>
+  items.reduce((arr, v) => (shouldBeHead(v) ? [v, ...arr] : [...arr, v]), [])
+
+const movePrimaryToHead = moveToHead(v => v.primary)
+
```
Crash--: This is something we should move outside of this file. Pretty sure it's already planned :D

src/components/ContactCard/ContactForm.jsx
```diff
@@ -77,7 +77,62 @@ const initialFieldValues = fields.reduce((initialValues, { name, isArray }) => {
...
+        cozy: cozyValue,
+        cozyLabel: cozyLabel,
+        company,
+        birthday,
+        note
+      }
+    }
+
+    return initialFieldValues
+  }
```
Crash--: Can we move this method outside of the component since it has nothing to do with the rendering, just formating.

src/components/ContactCard/ContactForm.spec.jsx
```diff
@@ -1,28 +1,102 @@
...
+      number: '+33 (2)0 90 00 54 04',
+      primary: true
+    },
+    {
+      number: '+33 6 77 11 22 33',
+      primary: false
+    }
+  ]
+}
+
```
Crash--: I know this is not the scope of this PR. But we should create a fixture file for our spec ?

src/components/ContactCard/ContactForm.spec.jsx
```diff
@@ -1,28 +1,102 @@
...
 describe('ContactForm', () => {
   it('should match snapshot', () => {
     const tree = renderer
       .create(
-        <I18n lang="en" dictRequire={() => ''}>
+        <I18n lang="en" dictRequire={dictRequire}>
```
Crash--: 👍

src/components/Modals/ContactCardMenu.jsx
```diff
@@ -0,0 +1,37 @@
...
+  </Menu>
+)
+ContactCardMenu.propTypes = {
+  deleteAction: PropTypes.shape({
+    label: PropTypes.string.isRequired,
+    action: PropTypes.func.isRequired
+  }).isRequired
+}
+
+export default ContactCardMenu
```
Crash--: <3.

src/components/Modals/ContactCardModal.jsx
```diff
@@ -76,24 +54,42 @@ class ContactCardModal extends React.Component {
...
+                        />
+                      </ModalHeader>
+                    )}
+                    renderBody={children => (
+                      <ModalContent>{children}</ModalContent>
+                    )}
+                  />
+                )}
+
+              {editMode && (
```
Crash--: && loaded ? I don't know if we can have some strange behavior here

src/components/Modals/ContactFormModal.jsx
```diff
@@ -22,19 +24,40 @@ const ContactFormModal = ({
...
+            const payload = {
+              _type: contact._type,
+              _id: contact._id,
+              _rev: contact._rev,
+              ...newContact
+            }
+            updateContact(payload).then(resp => {
+              // TODO: go back to view mode
+              console.log('Done: ', resp)
+            })
```
Crash--: async / await ? :D

src/components/ContactCard/ContactForm.jsx
cedricmessiant: in this case, the default export is the "connected" `ContactForm`, the Dumb one is for tests (and to avoid to have both default export and named export with the name of the file). I should/will use export / as for this actually

src/components/ContactCard/ContactForm.jsx
cedricmessiant: I agree but if I do this I should probably also move `formDataToContact` method outside too (rewrite it so it only convert data and add a `handleSubmit` method that calls this function and `onSubmit(contact)`). For the react-final-form forms, I usually used a folder for the form with the component and various utils in separate files (`load.js`, `save.js`, `validate.js`)

src/components/ContactCard/ContactForm.spec.jsx
cedricmessiant: you mean test data that can be used in various unit tests?

Crash--: yep

src/components/ContactCard/ContactForm.jsx
Crash--: Like the idea

cedricmessiant: @Crash-- you talk about the `save.js`/`load.js/`validate.js` files?
@y-lohse what do you think about it?

src/components/ContactCard/ContactForm.spec.jsx
cedricmessiant: I can move the contact I created in `src/helpers/testData.js` or somewhere else. It would be a starting point

src/components/ContactCard/ContactForm.jsx
cedricmessiant: I would like to but I can't find the right place for now :smile: Maybe a new file in `helpers/` I guess

src/components/ContactCard/ContactForm.jsx
y-lohse: 👍 for separating the purely logic-code into standalone files. And as long as the code is coupled to the forms, I think having them next to the components is good, so `save.js/load.js` works for me too!

cedricmessiant: In the end, I chose `contactToFormValues.js` and `formValuesToContact.js` because load and save suggest that the data is fetched and pushed to the server

src/components/ContactCard/ContactForm.jsx
cedricmessiant: moved to `contactToFormValues.js`

src/components/Modals/ContactCardModal.jsx
```diff
@@ -76,26 +54,47 @@ class ContactCardModal extends React.Component {
...
+                  />
+                )}
+
+              {editMode &&
+                fetchStatus === 'loaded' && (
+                  <ContactFormModal
+                    contact={contact}
+                    onClose={this.toggleEditMode}
+                    title={t('edit-contact')}
+                    createContact={() => {}}
```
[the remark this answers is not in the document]
cedricmessiant: I forgot to remove this, thanks

PR #425, opened 2019-01-22
fix: ImportBar alignment

Crash--:
before: 
before

after: 
after

### DIFF
--- a/src/components/ContactImportationModal/ContactImportationCard/index.jsx
+++ b/src/components/ContactImportationModal/ContactImportationCard/index.jsx
@@ -30,7 +30,13 @@ const ContactImportationCard = props => {
           <span className="importation-status">{t(statusKey(status))}</span>
         )
       )}
-      {progress && <progress value={progress.current} max={progress.total} />}
+      {progress && (
+        <progress
+          value={progress.current}
+          max={progress.total}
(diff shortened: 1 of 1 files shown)


Crash--:
[Contacts: Barre de progression d'import non centrée](https://trello.com/c/WrCUhaZw/1580-contacts-barre-de-progression-dimport-non-centr%C3%A9e)

src/components/ContactImportationModal/ContactImportationCard/index.jsx
```diff
@@ -30,7 +30,13 @@ const ContactImportationCard = props => {
           <span className="importation-status">{t(statusKey(status))}</span>
         )
       )}
-      {progress && <progress value={progress.current} max={progress.total} />}
+      {progress && (
+        <progress
+          value={progress.current}
+          max={progress.total}
+          className={'importation-progress'}
```
y-lohse: Do you now if this works too? 

```suggestion
          className={'u-m-auto u-mv-0'}
```

It would avoid adding special class

Crash--: Thank you very much! Didn't know about this one (mv-0)

PR #427, opened 2019-01-23
feat: Use new small cross and small plus icons

Crash--:
wait for https://github.com/cozy/cozy-ui/pull/818

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
   "homepage": "https://github.com/cozy/cozy-contacts#readme",
   "devDependencies": {
     "babel-preset-cozy-app": "1.3.0",
-    "cozy-scripts": "1.12.1",
+    "cozy-scripts": "1.11.0",
     "enzyme": "3.8.0",
     "enzyme-adapter-react-16": "1.7.1",
     "eslint-config-cozy-app": "1.1.8",
@@ -51,7 +51,7 @@
     "cozy-bar": "5.0.9",
     "cozy-client": "4.7.0",
     "cozy-flags": "1.2.11",
-    "cozy-ui": "18.4.1",
+    "cozy-ui": "18.12.0",
     "cozy-vcard": "0.2.18",
     "final-form": "4.10.0",
     "final-form-arrays": "1.1.0",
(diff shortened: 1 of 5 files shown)


Crash--:
[Contacts - formulaire création/édition - pictos](https://trello.com/c/IkFyodjT/1566-contacts-formulaire-cr%C3%A9ation-%C3%A9dition-pictos)

package.json
```diff
@@ -34,7 +34,7 @@
   "homepage": "https://github.com/cozy/cozy-contacts#readme",
   "devDependencies": {
     "babel-preset-cozy-app": "1.3.0",
-    "cozy-scripts": "1.12.1",
+    "cozy-scripts": "1.11.0",
```
y-lohse: Why are we downgrading cozy-scripts in this PR?

src/components/ContactCard/ContactForm/ContactFormField.jsx
```diff
@@ -82,7 +82,7 @@ class ContactFormField extends React.Component {
             onClick={() => this.addField(fields)}
             className="contact-form__meta-button contact-form__meta-button--add"
           >
-            <Icon icon={'plus'} color="white" width="10" height="10" />
+            <Icon icon={'plus-small'} color="white" width="10" height="10" />
```
y-lohse: For the plus icon above, we changed the size from 10 to 12, but not here. Is it normal?

package.json
Crash--: See the chan. renovate has updated CS during the night and we has an issue with this release. 

We expect a new release today / tomorrow.

src/components/ContactCard/ContactForm/ContactFormField.jsx
Crash--: Yep, it looks great that way ;).

package.json
y-lohse: Mmh alright, but renovate is going to autoupdate it again, and if the newest version isn't out by then we'll have the same problem.

src/components/ContactCard/ContactForm/ContactFormField.jsx
y-lohse: Best reason 😎

PR #442, opened 2019-02-04
fix: Update cozy-client

y-lohse:
Needs https://github.com/cozy/cozy-client/pull/355 + a new cozy-client version.

### DIFF
--- a/src/components/ContactCard/ContactGroups.jsx
+++ b/src/components/ContactCard/ContactGroups.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import flow from 'lodash/flow'
+import differenceBy from 'lodash/differenceBy'
 import { Query } from 'cozy-client'
 import { Alerter } from 'cozy-ui/transpiled/react'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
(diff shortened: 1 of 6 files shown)


y-lohse:
[(0.5) Contacts : cozy-client & relationships](https://trello.com/c/jv67vA73/1591-05-contacts-cozy-client-relationships)

src/components/ContactCard/ContactGroups.jsx
```diff
@@ -15,17 +16,17 @@ import SpinnerContact from '../Components/Spinner'
...
-      contact.groups.addById(groupToAdd._id)
-    })
-    this.props.updateContact(contact)
+    const currentGroups = contact.groups.data
+    const toAdd = differenceBy(nextGroups, currentGroups, '_id')
+    const toRemove = differenceBy(currentGroups, nextGroups, '_id')
+
+    if (toAdd.length > 0) contact.groups.addById(toAdd.map(({ _id }) => _id))
+    else if (toRemove.length > 0)
+      contact.groups.removeById(toRemove.map(({ _id }) => _id))
```
Crash--: you can have toAdd and toRemove at the same time, no?

Crash--: And maybe we should add a test with this case : 
- adding 1 group 
- removing one other

y-lohse: No we can't have that as far as I can tell — and actually it will break cozy-client if that happens. I still have to file a bug about that, but it's not one that can be fixed easily…

y-lohse: Added a comment highlighting why we can't do this right now 👍

PR #456, opened 2019-02-20
feat: Flag contacts with sources as trashed

y-lohse:
Contacts without sources are deleted right away, those with sources get a `trashed` flag and are ignored in the UI.

Two things to note:

- We need to use `_id` in the query as well, because we can't run a query only a non existing field.
- We're not using the cozy-client relationships to detect sources and instead rely on the metadata. This is because the google konnector doesn't generate the relationship at the moment. We should update this code once it's done, although it will continue to work.

### DIFF
--- a/src/connections/allContacts.spec.js
+++ b/src/connections/allContacts.spec.js
@@ -1,0 +1,63 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import withContactsMutations from './allContacts'
+import AppLike from '../tests/Applike'
+
+const DummyComponent = () => <span />
+const DummyComponentWithMutations = withContactsMutations(DummyComponent)
+
+describe('Cozy Contacts API', () => {
+  let testedComponent, client
+  beforeEach(() => {
+    client = {
+      create: jest.fn(),
+      save: jest.fn(),
+      destroy: jest.fn()
+    }
+
+    const root = (
+      <AppLike>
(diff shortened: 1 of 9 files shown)


y-lohse:
[(2) Synchro Google - suppression(s) Cozy->Google [2 pers.]](https://trello.com/c/7IXzLsRe/1535-2-synchro-google-suppressions-cozy-google-2-pers)

src/connections/allContacts.spec.js
```diff
@@ -0,0 +1,63 @@
...
+    expect(client.destroy).toHaveBeenCalledWith(contact)
+
+    const contactWithoutMetadata = {
+      _id: '456'
+    }
+    testedComponent.prop('deleteContact')(contactWithoutMetadata)
+    expect(client.destroy).toHaveBeenCalledWith(contactWithoutMetadata)
+  })
+
+  it('should flag a contact with sources as deleted', () => {
```
cedricmessiant: nit: maybe use `trashed` here also instead of `deleted`

y-lohse:
Sorry, just found a bug in this PR, BRB fixing it.

y-lohse:
Found the bug, waiting for https://github.com/cozy/cozy-client/pull/370 and a new cozy-client version.

y-lohse:
All good now!

PR #457, opened 2019-02-21
Display accounts linked to the contact in the contact card

cedricmessiant:
In the body of the contact card, renders the sources linked to the contact if any.

We could use `ContactFields` component for this but it would require adaptation because:
- icons and field values are computed from the field type
- fields icons are "forced" in grey

Also, it seems fair to have a distinction between fields and accounts.

What do you think about it?

### DIFF
--- a/src/components/ContactCard/ContactAccounts.jsx
+++ b/src/components/ContactCard/ContactAccounts.jsx
@@ -1,0 +1,56 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+import { Text, Caption } from 'cozy-ui/transpiled/react/Text'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+
+import IconGoogle from '../../assets/icons/connect-google.svg'
+
+const ACCOUNTS_MAPPING = {
+  'konnector-google': {
+    label: 'Google',
+    icon: IconGoogle
+  }
+}
+
+const ContactAccounts = ({ accounts, t }) => (
+  <div className="contact-accounts">
+    <h3 className="contact-fields-title">
+      {t('associated_accounts', {
+        smart_count: accounts.length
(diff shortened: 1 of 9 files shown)


cedricmessiant:
[(1) Fiche contact - vue sources connectées](https://trello.com/c/wMtyA3WH/1526-1-fiche-contact-vue-sources-connect%C3%A9es)

src/components/ContactCard/ContactAccounts.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import { translate } from 'cozy-ui/transpiled/react/I18n'
+import { Text, Caption } from 'cozy-ui/transpiled/react/Text'
+import Icon from 'cozy-ui/react/Icon'
```
Crash--: transpiled ?

src/components/ContactCard/ContactAccounts.jsx
```diff
@@ -0,0 +1,45 @@
...
+              </div>
+            </div>
+          </li>
+        )
+      })}
+    </ol>
+  </div>
+)
+
+export default translate()(ContactAccounts)
```
Crash--: proptypes ?

src/components/ContactCard/ContactAccounts.jsx
```diff
@@ -0,0 +1,45 @@
...
+      })}
+    </h3>
+    <ol className="contact-field-list">
+      {accounts.map(account => {
+        const { icon, label } = ACCOUNTS_MAPPING[account.type]
+        return (
+          <li key={account.id}>
+            <div className="contact-field">
+              <div className="contact-field-icon">
+                <Icon icon={icon} height="16" width="16" />
```
Crash--: I think 16 is already the default size (https://github.com/cozy/cozy-ui/blob/master/react/Icon/index.jsx#L6)

PR #473, opened 2019-03-13
feat: Better delete warning

y-lohse:
Sorry, a bit light on the tests — but adding snapshots for the delete modal requires a lot more effort than what we wanted to give this task.

### DIFF
--- a/src/components/layout/ContactsSelectionBar.jsx
+++ b/src/components/layout/ContactsSelectionBar.jsx
@@ -4,6 +4,7 @@ import { SelectionBar, Modal } from 'cozy-ui/react'
 import withSelection from '../Selection/selectionContainer'
 import withModalContainer from '../HOCs/withModal'
 import { translate } from 'cozy-ui/transpiled/react/I18n'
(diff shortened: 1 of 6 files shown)


y-lohse:
[Contacts - wording modale suppression](https://trello.com/c/xTGuyCzl/1754-contacts-wording-modale-suppression)

src/components/layout/ContactsSelectionBar.jsx
```diff
@@ -22,13 +23,28 @@ class ContactsSelectionBar extends Component {
         actions={{
           trash: {
             action: () => {
+              const contactsWithAccounts = selection.filter(
+                contact => getConnectedAccounts(contact).length > 0
+              )
+              const allContactsConnected =
+                contactsWithAccounts.length === selection.length
+              const someContactsConnected =
```
cedricmessiant: nit: I would have used `selection.every` and `selection.some`

y-lohse: Good point, I rarely think of those.

PR #479, opened 2019-03-15
Contact form improvements

cedricmessiant:
- Change label placeholder to label (TODO: change it in French translations)
- Field arrays: Replace add button on row by a unique button under all rows
- Display label field if it has a value (see https://codesandbox.io/s/lm4p3m92q for an example + doc about `Condition` field in react-final-form)
- If a field that has a label has been focused, its label does not disappear anymore
- use cozy-ui fields

### DIFF
--- a/src/components/Form/fields.jsx
+++ b/src/components/Form/fields.jsx
@@ -1,0 +1,27 @@
+import React from 'react'
+import Input from 'cozy-ui/transpiled/react/Input'
+import Textarea from 'cozy-ui/transpiled/react/Textarea'
+
+function adaptField(WrappedComponent) {
+  class AdaptField extends React.Component {
+    render() {
+      const {
+        input,
+        meta: { valid },
+        ...rest
+      } = this.props
+      return <WrappedComponent {...input} {...rest} data-valid={valid} />
+    }
+  }
+  AdaptField.displayName = `adaptField(${getDisplayName(WrappedComponent)})`
+  return AdaptField
+}
+
+function getDisplayName(WrappedComponent) {
+  return WrappedComponent.displayName || WrappedComponent.name || 'Component'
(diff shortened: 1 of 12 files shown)


cedricmessiant:
[(𝐒) Contacts - Form édition de contacts - comportements form](https://trello.com/c/G2bAYM2J/1729-%F0%9D%90%92-contacts-form-%C3%A9dition-de-contacts-comportements-form)

cedricmessiant:
Edit: react-final-form Condition doc/example link

src/components/Form/fields.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import Input from 'cozy-ui/transpiled/react/Input'
+import Textarea from 'cozy-ui/transpiled/react/Textarea'
+
+function adaptField(WrappedComponent) {
+  class AdaptField extends React.Component {
+    render() {
+      const {
+        input,
+        meta: { valid },
```
Crash--: Can we use custom attributes? Something like `data-valid` ?

cedricmessiant: I can change it here: `<WrappedComponent {...input} {...rest} data-valid={valid} />`. Note that we don't use `react-final-form`'s validation so `meta.valid` is always `true` and `cozy-ui` does not take either `valid` or `data-valid` prop into account for now.

Crash--: That is exactly my point: I don't think having a `valid` attribute in the generated DOM is a good idea since `valid` is not a valide attribute : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Attributes

cedricmessiant: done, thanks!

PR #480, opened 2019-03-15
fix: Update cozy-ui

y-lohse:
The snapshot diff is a bit hard to read, but AFAIK there's only a wrapping div that has been removed.

### DIFF
--- a/src/styles/index.styl
+++ b/src/styles/index.styl
@@ -1,4 +1,4 @@
-@require 'cozy-ui/build'
+@import '~cozy-ui/dist/cozy-ui.min.css'
 
 @require './contacts.styl'
 @require './contactCard.styl'
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "cozy-bar": "6.10.0",
     "cozy-client": "6.9.0",
     "cozy-flags": "1.2.11",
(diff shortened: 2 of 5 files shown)


src/styles/index.styl
```diff
@@ -1,4 +1,4 @@
-@require 'cozy-ui/build'
+@import '~cozy-ui/dist/cozy-ui.min.css'
```
y-lohse: Taken from the updated template https://github.com/CPatchane/create-cozy-app/blob/master/packages/cozy-scripts/template/app/src/styles/index.css#L2

Not ideal though as I think we have duplicated styles.

cedricmessiant: Why duplicated styles? Maybe we could compare the bundle sizes?

GoOz: I'm absolutely 100% positive we're having duplicated styles with this but, indeed, that's the only way to have erverything working so far.

y-lohse: Yep the bundle size probably exploded, but the alternate solution is not clear to me — see https://github.com/cozy/cozy-ui/issues/866

Crash--:
@y-lohse Any news on this PR?

y-lohse:
Yeah we can close it, the same has been done in another PR.

PR #483, opened 2019-03-20
feat: revamp of contact card with utility classes

GoOz:
Remade & fixed the contact card with utility classes so it can be used elsewhere.

There's a lot to digest. Sorry.

~~⚠️ Waiting for an update of Cozy-UI~~ ✅


GoOz:
[(𝐌) Contacts - Révision UI de la fiche contact](https://trello.com/c/yD0TpuOh/1115-%F0%9D%90%8C-contacts-r%C3%A9vision-ui-de-la-fiche-contact)

GoOz:
[(𝐌) Contacts - Fiche contact révision UI](https://trello.com/c/0KDJUugd/1728-%F0%9D%90%8C-contacts-fiche-contact-r%C3%A9vision-ui)

src/components/Components/FieldsTitle.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+const FieldsTitle = ({ title }) => (
```
gregorylegarec: If I understand well, `FieldsTitle` here is a title for a group a fields. Maybe `FieldsetTitle` could be more accurate ? What do you think ?

src/components/Components/FieldsList.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+
+const FieldsList = ({ children }) => (
```
gregorylegarec: Maybe `Fieldset` could be a more accurate name ?

I personnaly prefer [_ItemList_ over _ItemsList_](https://stackoverflow.com/questions/9784720/items-list-or-item-list), but it's a matter of personal preferences. Maybe we could defined a styleguide for naming our list.

src/components/ContactCard/__snapshots__/ContactCard.spec.jsx.snap
```diff
@@ -239,12 +239,12 @@ Array [
 `;
 
 exports[`ContactCard should match snapshot with accounts 1`] = `
-<div>
+Array [
```
gregorylegarec: We're passing from a single node to a node array, I guess it is expected, but why ?

src/components/Components/FieldsTitle.jsx
GoOz: Well, yeah. I honestly just took the previous className and transformed it into a component's name so it can be less confusing. But your comment is legit.

src/components/Components/FieldsTitle.jsx
```diff
@@ -0,0 +1,12 @@
...
+  <h3 className="u-title-h2 u-mt-1-half u-mb-1">{title}</h3>
+)
+
+FieldsTitle.propTypes = {
+  title: PropTypes.string
+}
+
+export default FieldsTitle
```
y-lohse: @Crash-- Would this be a good candidate to start using `React.memo`? Doesn't have to be in this PR, more asking out of curiosity.

src/styles/contactGroupCreation.styl
```diff
@@ -15,16 +15,17 @@
...
 /*since we keep the existing style of
-react-select for the menu and we only 
-want to override with & zindex... important 
+react-select for the menu and we only
+want to override with & zindex... important
 is important */
 .react-select__menu
     width rem(384) !important
     z-index 100000 !important
+    right 0
```
y-lohse: Why do we have to add this now? If it's for a good reason, we should update the comment above as well.

src/components/ContactCard/ContactAccounts.jsx
```diff
@@ -14,23 +16,21 @@ const ACCOUNTS_MAPPING = {
...
 
 const ContactAccounts = ({ accounts, t }) => (
-  <div className="contact-accounts">
-    <h3 className="contact-fields-title">
-      {t('associated_accounts', {
-        smart_count: accounts.length
-      })}
-    </h3>
-    <ol className="contact-field-list">
+  <div>
```
Crash--: do we really need this `div` since we can use `<>`?

src/components/ContactCard/ContactIdentity.jsx
```diff
@@ -1,23 +1,26 @@
 import React from 'react'
 import { Avatar } from 'cozy-ui/transpiled/react'
+import withBreakpoints from 'cozy-ui/react/helpers/withBreakpoints'
```
Crash--: transpiled ?

src/components/ContactCard/__snapshots__/ContactCard.spec.jsx.snap
Crash--: This is how enzyme-to-json renders a component that has a fragment root 

https://github.com/adriantoine/enzyme-to-json/pull/90/files#diff-7223bdb24283e1d40122de3c584939c3R136

src/components/ContactGroups/ContactGroupList.jsx
```diff
@@ -0,0 +1,38 @@
...
+      .filter(value => value)
+
+    return (
+      <div className="u-mb-2">
+        {title && <FieldsTitle title={title} />}
+        <ol className="u-nolist u-m-0 u-p-0">
+          {userGroups.map(group => (
+            <li
+              key={group._id}
+              className="u-dib u-slateGrey u-fz-small u-p-half u-mr-half u-w-auto u-maw-4 u-bg-paleGrey u-ellipsis"
```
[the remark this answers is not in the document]
GoOz: Sorry what? 
You mean this could be an external component just like `<FieldsTitle />` & `<FieldsList />` ?

src/styles/contactGroupCreation.styl
GoOz: As the select goes to te right of the modal, we want the dropdown to start from the right to the left.
This was a design request. I'll update the comment.

src/components/Components/FieldsTitle.jsx
Crash--: It can be, yes!

src/components/ContactGroups/ContactGroupList.jsx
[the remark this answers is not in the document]
y-lohse: Yes !

src/components/ContactCard/ContactCard.jsx
```diff
@@ -27,17 +32,32 @@ const ContactCard = ({ title, contact, renderHeader, renderBody }) => {
...
         <>
-          <ContactFields fields={normalizedFields} title={title} />
+          <Query query={allGroupsQuery}>
+            {({ data: allGroups, fetchStatus }) => {
+              if (fetchStatus === 'loaded') {
+                return (
+                  <ContactGroupsList
+                    contact={contact}
+                    allGroups={allGroups}
+                    title={t('contact_group')}
```
Crash--: Since we are displaying `t('contact_group')` even if the contact doesn't have any group, can we do something like : 

```jsx 
 {({ data: allGroups, fetchStatus }) => {
return (
<ContactGroupsList
contact={contact}
allGroups={allGroups}
title={t('contact_group')}
status={fetchStatus}
/>
```

And in the ContactGroupsList https://github.com/cozy/cozy-contacts/pull/483/files#diff-e89cc739ce1afe47f19587017466807bR18 if (status === 'loaded') { <ul>... } else { <Spinner /> } ? 

ATM : 
- the current displayed spinner is small, smaller than `contact_group` title. So on a slow network, we'll have a visual glitch 

- Are we sure about the "no group" status? Currently we'll display "Groups" with blanks and below `Contact Informations` : 
  - adding "no group assigned yet" can be a good solution?

src/components/ContactCard/ContactCard.jsx
```diff
@@ -27,17 +32,32 @@ const ContactCard = ({ title, contact, renderHeader, renderBody }) => {
   const activeContactAccounts = getConnectedAccounts(contact)
 
   return (
-    <div>
+    <>
       {renderHeader(<ContactIdentity contact={contact} />)}
       {renderBody(
         <>
-          <ContactFields fields={normalizedFields} title={title} />
+          <Query query={allGroupsQuery}>
```
Crash--: J'ai un problème avec ça en fait, car ça veut dire que pour afficher un contact, on va faire 2 fois la requête pour récupérer tous les groupes : 

- Une fois dans le renderHeader via `ContactGroups` 
- Une fois dans le renderBody via la `Query` ci-dessus 

Dans l'ideal, il faudrait : 
- Récupérer la `allGroupsQuery` dans le <ContactCardModal> 
- Passer en props ces groups récupérés au Header et au Body

Crash--: @GoOz Let's focus on the UI part (see https://github.com/cozy/cozy-contacts/pull/483#discussion_r268779339), I'll take care of this one (see #484)

src/components/Components/GroupItem.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+const GroupItem = ({ children, key }) => (
+  <li
+    key={key}
```
y-lohse: Nit:  I might be mistaken, but I don't think you need the `key` prop here. It's only important to have it on the component that is created within a loop, not so much in the render itself.

(for more infos: https://overreacted.io/react-as-a-ui-runtime/#lists)

Crash--:
I'm merging this PR. Will make some change in #484

PR #484, opened 2019-03-25
fix: Use only one Query to get allGroups

Crash--:
This is the first try to use only one query to get `allGroups`.

First time ever I use `queryConnect` 💪

### DIFF
--- a/src/components/ContactCard/ContactCard.spec.jsx
+++ b/src/components/ContactCard/ContactCard.spec.jsx
@@ -8,7 +8,7 @@ import {
 } from '../../helpers/doctypes'
 
 describe('ContactCard', () => {
-  it('should match snapshot', () => {
+  it('should match snapshot when groups are available but no one is linked to the specified contact', () => {
     const contact = {
       _id: 'c6899688-6cc6-4ffb-82d4-ab9f9b82c582',
       _rev: '1-9368a4f2e467c449f4a1f5171a784aa8',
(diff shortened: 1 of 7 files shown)


Crash--:
This PR is ready to be reviewed. 

I can't use queryConnect since it's buggy (will attach the GH issue pretty soon)

Crash--:
[(𝐌) Contacts - Fiche contact révision UI](https://trello.com/c/0KDJUugd/1728-%F0%9D%90%8C-contacts-fiche-contact-r%C3%A9vision-ui)

src/components/Modals/ContactCardModal.jsx
```diff
@@ -45,84 +45,113 @@ class ContactCardModal extends React.Component {
...
-                        </div>
-                      </ModalHeader>
-                    )}
-                    renderBody={children => (
-                      <ModalContent>{children}</ModalContent>
-                    )}
-                  />
-                )}
+      <Modal into="body" dismissAction={onClose} size="xlarge" mobileFullscreen>
+        <Query query={client => client.get(DOCTYPE_CONTACTS, id)}>
```
cedricmessiant: why did you switch `<Modal>` and `<Query>`?

src/components/ContactCard/ContactCard.spec.jsx
```diff
@@ -8,7 +8,7 @@ import {
 } from '../../helpers/doctypes'
 
 describe('ContactCard', () => {
-  it('should match snapshot', () => {
+  it('should match snapshot without groups', () => {
```
y-lohse: Why `without groups`? There are groups in this snapshot..?

src/components/Modals/ContactCardModal.jsx
Crash--: Because I want to open the Modal before since now I need to wait the resolution of 2 requests

src/components/ContactCard/ContactCard.spec.jsx
Crash--: Oh you're right, it should probably something like: 

it should match snapshot when groups are available but no one is linked to the specified contact

PR #485, opened 2019-03-26
fix: Add onDismiss function

Crash--:
To remove proptypes warning


Crash--:
Let's see https://github.com/cozy/cozy-ui/pull/883

Crash--:
Will upgrade to the latest UI, thanks!

PR #490, opened 2019-04-09
fix: cozy url is an array, not a string

cedricmessiant:
If cozy url is empty, it should be an empty array, not an empty string in contact doctype.


cedricmessiant:
[champ 'cozy' dans le doctype contact](https://trello.com/c/5pr7FR8z/1804-champ-cozy-dans-le-doctype-contact)

PR #498, opened 2019-04-29
feat: Add migration script for contacts

cedricmessiant:
See comment at the top of `migrateContacts.js` for what it does.

A few questions:
- I should probably rename `migrateContacts.js` but can't really find a good name. Maybe at least add a version `contactsMigration1.js`?
- I could add a test for `run`/`doMigrations` but I'm not sure that it worths it.
- it introduces a dependency to ACH. Maybe it should at least be a `devDependency`. @ptbrowne how do you handle this for external scripts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "devDependencies": {
     "babel-core": "7.0.0-bridge.0",
     "babel-preset-cozy-app": "1.5.1",
+    "cozy-ach": "1.19.0",
     "cozy-scripts": "1.13.0",
     "enzyme": "3.9.0",
     "enzyme-adapter-react-16": "1.12.1",
--- a/scripts/20190429_migrateContacts.js
+++ b/scripts/20190429_migrateContacts.js
@@ -1,0 +1,165 @@
+/*
+ * This migration script is meant to be used with ACH. It handles the following cases:
+ *
+ * - ghosts contacts i.e contacts with sources that were deleted since Contacts 0.8.0
+ * and the related google connector was uninstalled (connector prior to 2.0.5)
+ * - contacts with an empty string in cozy attribute (contacts created with Contacts 0.8.1 with an empty cozy)
+ * - old contacts with a string in cozy or email attributes
+*/
+const get = require('lodash/get')
+
+const { mkAPI, utils } = require('cozy-ach')
+
+const DOCTYPE_CONTACTS = 'io.cozy.contacts'
+const DOCTYPE_CONTACTS_ACCOUNT = 'io.cozy.contacts.accounts'
+
+function fixCozy(contact) {
+  if (contact.cozy === '') {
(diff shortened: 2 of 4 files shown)


cedricmessiant:
[Migration des contacts](https://trello.com/c/m8TtEoQT/1869-migration-des-contacts)

package.json
```diff
@@ -51,6 +51,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
+    "cozy-ach": "^1.19.0",
```
Crash--: no ^

cedricmessiant: oh, sorry. Maybe we could add a pre-commit hook for this...

y-lohse: We can, but honestly I don't think we need to change this in PRs. Renovate does it automatically for us once merged.
Another trick is to use `yarn add cozy-ach -E` when adding the dependency.

y-lohse:
I would say `devDependency ` is the way to go for this. Definitely not a regular dependency though.

scripts/migrateContacts.js
```diff
@@ -0,0 +1,168 @@
...
+
+  const fixedContacts = contacts
+    .map(contact => {
+      const cozyResult = fixCozy(contact)
+      if (cozyResult.cozyFixed) result.cozy++
+      if (cozyResult.emptyCozyFixed) result.cozyEmpty++
+      const { emailFixed, contact: fixedContact } = fixEmail(cozyResult.contact)
+      if (emailFixed) result.email++
+
+      if (cozyResult.cozyFixed || emailFixed) {
```
y-lohse: And if `cozyResult .emptyCozyFixed` ?

scripts/migrateContacts.js
```diff
@@ -0,0 +1,168 @@
...
+
+  if (dryRun) {
+    logWithInstance(
+      `Would process ${contacts.length} contacts: \n[cozy]: ${
+        result.cozy
+      } would be fixed.\n[empty cozy]: ${
+        result.cozyEmpty
+      } would be fixed.\n[email]: ${result.email} would be fixed.\n[ghosts]: ${
+        ghosts.length
+      } would be deleted.`
```
y-lohse: Not a problem, but the line breaks are weird here? It would be easier to read if they matched the `\n` I think.

y-lohse:
> I should probably rename migrateContacts.js but can't really find a good name. Maybe at least add a version contactsMigration1.js?

Maybe we can prefix it with the date ?

> I could add a test for run/doMigrations but I'm not sure that it worths it.

I think there's a bug in the `doMigrations` logic (see comment above)… so it might be worth adding a test on that :)

cedricmessiant:
Added a test, thanks @y-lohse there was a huge bug in the `doMigrations` script!

PR #500, opened 2019-05-03
Feat/new contact form

GoOz:
- [x] Move Edit button to the Menu dropdown
- [x] new Contacts Form design
Screenshot 2019-05-03 19 01 11

~~⚠️ Can't save or cancel edition yet, waiting for inputs from @cedricmessiant~~ ✅

~~⚠️ Needs an update of Cozy-UI when [#958](https://github.com/cozy/cozy-ui/pull/958) is merged.~~ ✅

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "cozy-bar": "6.10.0",
     "cozy-client": "6.9.0",
     "cozy-flags": "1.2.11",
-    "cozy-ui": "19.16.0",
+    "cozy-ui": "19.31.0",
     "cozy-vcard": "0.2.18",
     "final-form": "4.10.0",
     "final-form-arrays": "1.1.0",
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "cozy-bar": "6.10.0",
     "cozy-client": "6.9.0",
     "cozy-flags": "1.2.11",
-    "cozy-ui": "19.31.0",
+    "cozy-ui": "19.32.2",
     "cozy-vcard": "0.2.18",
     "final-form": "4.10.0",
     "final-form-arrays": "1.1.0",
--- a/src/components/ContactCard/ContactFieldInput.jsx
+++ b/src/components/ContactCard/ContactFieldInput.jsx
@@ -1,7 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Field } from 'react-final-form'
-import classnames from 'classnames'
 
 import HasValueCondition from '../Form/HasValueCondition'
 import { AdaptedInput, AdaptedTextarea } from '../Form/fields'
@@ -11,26 +10,12 @@ const getInputComponent = inputType =>
 
 class ContactFieldInput extends React.Component {
   state = {
-    hasBeenFocused: false,
-    hasFocus: false
+    hasBeenFocused: false
   }
 
   onFocus = () => {
     this.setState({
-      hasBeenFocused: true,
-      hasFocus: true
-    })
-  }
-
-  onSecondaryInputBlur = () => {
-    this.setState({
-      hasFocus: false
-    })
-  }
-
-  onMainInputBlur = () => {
-    this.setState({
-      hasFocus: false
+      hasBeenFocused: true
     })
   }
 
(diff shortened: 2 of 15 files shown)


src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -11,26 +10,12 @@ const getInputComponent = inputType =>
...
   state = {
-    hasBeenFocused: false,
-    hasFocus: false
+    hasBeenFocused: false
   }
 
   onFocus = () => {
     this.setState({
-      hasBeenFocused: true,
-      hasFocus: true
```
GoOz: It doesn't seem that the `hasFocus` prop is relevant anymore. Not quite sure though.

Crash--:
- Screenshots to fix.
- After the UI upgrade, we need to double check the display of the "Info" component (when we don't have any contacts since this component has been edited a lot)

cedricmessiant:
I think @Crash-- mean snapshots. I also saw errors in the js console (missing `htmlFor` prop and modal without label, you can ignore the `onSelectDisabled` one)

src/components/ContactCard/ContactFieldInput.jsx
cedricmessiant: it seems that it was there only to style the input. Seems fine for me

src/components/ContactCard/ContactForm/ContactFormField.jsx
```diff
@@ -9,17 +11,22 @@ class ContactFormField extends React.Component {
   render() {
     const { name, icon, label, isArray, renderInput, t } = this.props
     return (
-      <div className="contact-form__field">
-        <label className="contact-form__label">
+      <div className="contact-form-field">
+        <Label
+          block={false}
```
cedricmessiant: react complains that `htmlFor` prop is missing

src/components/Modals/ContactFormModal.jsx
```diff
@@ -43,9 +45,25 @@ const ContactFormModal = ({
...
+        type="button"
+        theme="secondary"
+        label="cancel"
+        onClick={onClose}
+      />
+      <Button
+        type="submit"
+        label="save"
+        onClick={() =>
+          document
```
cedricmessiant: maybe I should add a comment here to explain why we need this?

GoOz:
[(𝐒) Contacts - Form édition de contacts - révision UI](https://trello.com/c/pUe8ei8H/1730-%F0%9D%90%92-contacts-form-%C3%A9dition-de-contacts-r%C3%A9vision-ui)

src/components/Modals/ContactFormModal.jsx
Crash--: Yep, I want an explanation about this direct dom manipulation. Because there is no way that we merge that ^^

cedricmessiant: Ok so I'll explain it here first :-) 
We need to be able to move the form buttons outside of the react-final-form because we need them in the modal footer so that they are sticky.

There are two ways to do this according to RFF documentation / examples : https://codesandbox.io/s/1y7noyrlmq Either we use a closure (I wanted to do this at first but we need a unique component with the modal and the form so... no) either we get the form id and submit it like I did here. Another solution would be that the form contains the modal components but I don't think it is a good idea. So it's not ideal but like this the form itself is still isolated and we will be able to use it for the myself page.

cedricmessiant: now we use `_submitContactForm` variable for this

Crash--:
@cedricmessiant @GoOz Still some snapshots issues

GoOz:
I'm gonna update the UI deps & look at the tests.

GoOz:
UI updated to 19.32.2

package.json
```diff
@@ -54,7 +54,7 @@
     "cozy-bar": "6.10.0",
     "cozy-client": "6.9.0",
     "cozy-flags": "1.2.11",
-    "cozy-ui": "19.16.0",
+    "cozy-ui": "19.31.0",
```
Crash--: 19.32.2 no ?

PR #503, opened 2019-05-14
fix: Translate save/cancel buttons in form

cedricmessiant:


cedricmessiant:
[(𝐒) Contacts - Form édition de contacts - révision UI](https://trello.com/c/pUe8ei8H/1730-%F0%9D%90%92-contacts-form-%C3%A9dition-de-contacts-r%C3%A9vision-ui)

PR #505, opened 2019-05-15
Pass our own cozy client to bar

cedricmessiant:
Make the bar use our own `CozyClient` (fixes an issue that appeared since cozy-bar was upgraded to 6.25.4)

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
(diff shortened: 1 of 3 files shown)


Crash--:
Since https://github.com/cozy/cozy-bar/releases/tag/v6.14.0 ;) 

But this commit should work https://github.com/cozy/cozy-bar/commit/ba811bd 

Let's merge it anyway

PR #596, opened 2019-12-02
Fix/call order

Crash--:
- Add permissions needed for the cozy-bar
- Fix scheduler version in order to work well with hooks
- Fix call order to setStore to cozy-client

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -48,6 +48,11 @@
     "contactsAccounts": {
       "type": "io.cozy.contacts.accounts",
       "verbs": ["ALL"]
+    },
+    "settings": {
+      "description": "Required by the cozy-bar to display Claudy and know which applications are coming soon",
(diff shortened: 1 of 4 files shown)


Crash--:
> So to fix the call order, we need to move the initCozyBar call and wrap it in a set timeout? It's… really strange?

I really don't know why I've to wrap in a setTimeout((),0). I'll try without HMR to see if it comes from there, but yeah the real solution is to call initCozyBar after our setStore()

y-lohse:
Ok well, the rest LGTM. If we can remove the setTimeout then great, otherwise we'll have to live with it I suppose :)

ptbrowne:
Could you add a test checking that the initiliasation works correctly ? It would prevent a zealous refactorer from removing it.

nono:
Can we merge this PR?

PR #620, opened 2020-04-15
feat: Upgrade cozy-client and cozy-ui

Crash--:
- Upgrade Cozy-Client to get perf boost on large request 
- Upgrade Cozy-UI since cozy-client needs it 
- Change import path to fix issue 
- Add mui/core since ActionMenu is based on it now


src/components/Raccourcis.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+
+const goTo = lettre => {
+  const url = new URL(window.location.href)
+  url.hash = lettre
+  window.location.href = url
+}
+const Raccourcis = () => {
```
y-lohse, line 8: I don't think this was meant to be part of the PR, was it? :)

PR #628, opened 2020-06-11
Fix/delete contact from ContactCardModal

Crash--:
We can't delete a contact from its ContactCardModal since we don't display the Modal even if we should... 

- [x] Remove those snapshots and use testing lib to test the component's behavior
- [x] Use already existing fixtures
- [x] Move fetch & spinner outside of the DumbComp


src/components/Modals/ContactCardModal.jsx
```diff
@@ -62,84 +62,22 @@ class ContactCardModal extends React.Component {
...
-                        secondaryText={t('cancel')}
-                        secondaryAction={this.toggleConfirmDeleteModal}
-                        dismissAction={this.toggleConfirmDeleteModal}
-                      />
-                    )
-                  }
+                  return (
+                    <DumpContactCardModal
+                      fetchStatusContact={fetchStatusContact}
+                      allGroupsContactStatus={allGroupsContactStatus}
```
ptbrowne, line 68: AMHA we should let the componnet creating the Query be responsible to show the spinner if the data is not loaded. This would remove the need of the fetchStatusContact (which could maybe be named contactFetchStatus) property and would keep the query responsibility in a single component. This would also make DumbContactCardModal (careful, the component is named DumpContactCardModal here; dump -> dumb) easier to test as the test would not have to focus on the query side.

Crash--: Complètement d'accord. J'y suis juste allé comme un vrai bouché en déplaçant tout le composant dans un premier temps. Je voulais aussi le simplifier avec des useQuery() qui le rendrait bien plus lisible, mais t'as raison pour les fetch c'est rapide à faire.

src/components/Modals/ContactCardModal.spec.jsx
```diff
@@ -0,0 +1,97 @@
...
+    })
+
+    const jsx = (
+      <AppLike client={client}>
+        <DumbContactCardModal {...props} />
+      </AppLike>
+    )
+
+    const { getByText } = render(jsx)
+    expect(getByText('delete-confirmation.title'))
```
y-lohse, line 77: Nit: here we could also test that the other modal is also displayed (the one with the contact information).

JF-Cozy:
> I see you ended up regrouping the delete confirmation modal with the contact card modal, is it purely for ease of testing or was there another reason?
> 
> The changes LGTM! Weirdly though it seems Travis CI is not running on this branch..?

Yes it's for ease of testing

src/components/Modals/ContactCardModal.spec.jsx
JF-Cozy: Done here https://github.com/cozy/cozy-contacts/pull/628/commits/0fce4c4a5a737a1f9969d5090ec51f6a6c0e3a7a

PR #657, opened 2020-08-04
feat: Use cozy-client to create specifics Contacts doctypes

JF-Cozy:
- Utilisation de cozy-client pour l'ajout des nouveaux attributs de doctype contact lors de l'ajout/modification d'un contact via le formulaire.
- Refacto de ContactCard pour sortir l'algo de normalisation des champs et le mettre dans le helper
- Les nouveaux attributs n'apparaissent pas dans la fiche d'un contact à sa consultation

C'est la version AVANT mise en place du service


Crash--:
It still WIP since it needs a merge on cozy-client. Also, you should not use the `merge` label for a PR waiting for an other merge, it's dangerous

src/components/ContactCard/ContactForm/formValuesToContact.spec.js
```diff
@@ -66,6 +66,11 @@ describe('formValuesToContact', () => {
         groups: {
           data: []
         }
+      },
+      displayName: 'John Doe',
+      indexes: {
+        byFamilyNameGivenNameEmailCozyUrl:
+          'DoeJohnjohn.doe@cozycloud.cchttps://johndoe.mycozy.cloud'
```
Crash--, line 73: I'm not sure that this is what we want at the end... 

We should probably remove the `http(s)://` from the cozy-url.

src/helpers/contacts.spec.js
```diff
@@ -157,3 +158,73 @@ describe('getConnectedAccounts', () => {
...
+      ],
+      name: { familyName: 'White', givenName: 'Rikki' },
+      fullname: 'Rikki White',
+      phone: [
+        { number: '+33 (1)9 14 02 28 31', primary: true },
+        { number: '+33 (2)3 99 53 65 21', primary: false }
+      ],
+      unknownField: 'unknownField',
+      emptyField: ''
+    }
```
Crash--, line 195: There is no `Contact Fixture` we can reuse? Instead of creating a new one? If we're adding new attributes to the contact, we'll need to upgrade all our "faked contact"

JF-Cozy:
> Nice work, LGTM.
> 
> For the record, ideally you fix/update the tests in the same commit that does the actual change, since technically the build should still be passing on all commits. That being said it's not always possible to follow that rule and it's not a big deal, just pointing it out for next time :)

Ok j'avais compris l'inverse, qu'il fallait séparer les commit de test des commit de 'code'. Merci de la précision :)

PR #661, opened 2020-08-19
Refactor: ContactCardModal to use func comp and useQuery

JF-Cozy:
This PR is a part [1] of multiple PR about the new sort logic in the application.

This refactor prevents problems of using `useQuery` in `App`. Otherwise It triggers errors when using `useQuery` in `App` and `Query` here in `ContactCardModal`.

- Update `react-redux` > 7.1.0 to be able to use `useQuery` hook
- Update `cozy-client` and fix some tests
- (*)Add `react-redux-test` (`react-redux` not updated) to prevent broken tests and `Enzyme` incompatibility
- Add some queries and update `ContactCardModal` in functional component using `useQuery`

(*) Using older version of `react-redux` is the best solution I found, I didn't success to make tests working in other way (like using `mount` of `Enzyme` or `react-test-renderer`, or without Provider). The problem is that `.dive()` doesn't work anymore and even if the right component is selected, the `.update()` doesn't do anything
See https://github.com/enzymejs/enzyme/issues/2202 , https://github.com/enzymejs/enzyme/issues/2302


src/targets/browser/index.jsx
```diff
@@ -14,13 +15,15 @@ import manifest from '../../../manifest.webapp'
...
 const RootApp = props => (
-  <CozyProvider client={props.client} store={props.store}>
-    <I18n lang={props.lang} polyglot={props.polyglot}>
-      <MuiCozyTheme>
-        <HotedApp />
-      </MuiCozyTheme>
-    </I18n>
-  </CozyProvider>
+  <Provider store={props.store}>
+    <CozyProvider client={props.client} store={props.store}>
```
JF-Cozy, line 19: I had to put `store={props.store}` on `Provider` but also on `CozyProvider`. Am I missing something or is it ok ?

ptbrowne: It's ok.

src/helpers/queries.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+}
+export const queryAllGroups = {
+  definition: () =>
+    Q('io.cozy.contacts.groups')
+      .where({
+        trashed: { $exists: false }
+      })
+      .sortBy([{ name: 'asc' }])
+      .indexFields(['name']),
```
ptbrowne, line 17: 1. Why do we need the `indexFields` here ? I thought that cozy-client would automatically index the fields in the sortBy ?

2. Here, if a group has { trashed: false }, it will not be retrieved, is it expected ?

src/helpers/queries.js
```diff
@@ -0,0 +1,21 @@
+import { Q } from 'cozy-client'
+import { DOCTYPE_CONTACTS } from './doctypes'
+
+export const queryContactById = {
+  definition: id => () => Q(DOCTYPE_CONTACTS).getById(id),
+  options: {
+    as: 'contactById'
```
ptbrowne, line 7: The name should contain the id, otherwise we'll reuse a wrong result if we change the id and add a fetch policy.

src/helpers/queries.js
ptbrowne: Adding a fetchPolicy would be good here.

ptbrowne (commented):
Looking good, I have some suggestions and questions though.

src/helpers/queries.js
JF-Cozy: 1. We don't need `indexFields`, I know now that it's not necessary 👍 
2. Yes it is

ptbrowne: 2. It seems counter-intuitive. Is this design decision discussed somewhere ?

Crash--: https://github.com/cozy/cozy-doctypes/blob/master/docs/io.cozy.contacts.md => `trashed?: {boolean} is true if the group is marked for removal and will be deleted soon` is clear enough? 

But even with that, I'm wondering why we ask this question? We want to display the `groups` the user have created in order to be able to set them on a contact or to display a list to filter the contacts' list. It makes sense to not fetch the deleted ones, no?

JF-Cozy: > Adding a fetchPolicy would be good here.

I tried but the fetch status of the request stay still in 'loading' during the time of the fetch policy. Expl: with the fetchPolicy at 5sec, the fetch status is loaded at T0, remain in 'loading' during T1 to T5, then works again at T5+ (if we trigger the request at T0, between T1 and T5, and after T5).

I looked at https://docs.cozy.io/en/cozy-client/api/cozy-client/#fetchpolicies and https://docs.cozy.io/en/cozy-client/react-integration/, get some examples from Banks and Drive to find some clues, but I think there is a problem with the store maybe ?

Crash--: Yep, the fetchStatus will still be loading, but now : 
- you have a lastFetch attribute in the query result https://github.com/cozy/cozy-banks/blob/master/src/ducks/client/utils.js#L16-L18
- https://github.com/cozy/cozy-banks/blob/40338ac6133c8b946ea60f6c3aece3dbc264cf1a/src/ducks/brandDictionary/selectors.js#L10-L12

ptbrowne: > It makes sense to not fetch the deleted ones, no?

My question is for the groups with { trashed: false }, not the ones with { trashed: true } ;)

JF-Cozy: Oh ok I saw and tested this lastFetch but didn't ring any bells :( Probably didn't understand how to use it. Thanks for the tips

ptbrowne: FYI, the fetchStatus still in loading is a bug but yes you can circumvent this with lastFetch.

src/helpers/queries.js
```diff
@@ -0,0 +1,21 @@
+import { Q } from 'cozy-client'
+import { DOCTYPE_CONTACTS } from './doctypes'
+
+export const queryContactById = {
+  definition: id => () => Q(DOCTYPE_CONTACTS).getById(id),
+  options: id => ({
+    as: `contactById-${id}`
+  })
+}
```
Crash--, line 9: why not passing the id directly to `queryContactById` and then using it queryContactById(id).definition ?

src/targets/browser/index.jsx
```diff
@@ -1,26 +1,30 @@
...
 import { I18n, initTranslation } from 'cozy-ui/transpiled/react/I18n'
+import MuiCozyTheme from 'cozy-ui/transpiled/react/MuiCozyTheme'
+
 import App from 'components/App'
 import configureStore from 'store/configureStore'
-import { hot } from 'react-hot-loader'
 import { schema } from '../../helpers/doctypes'
 import manifest from '../../../manifest.webapp'
-import MuiCozyTheme from 'cozy-ui/transpiled/react/MuiCozyTheme'
+import 'styles'
```
Crash--, line 16:
```suggestion
import 'styles.styl'
```

Crash--:
> (*)Add react-redux-test (react-redux not updated) to prevent broken tests and Enzyme incompatibility

Can you explain what was wrong with the latest react-redux? Is there any existing issue on Enzyme / Redux repo?

src/helpers/queries.js
Crash--: > My question is for the groups with { trashed: false }, not the ones with { trashed: true } ;)

Holy sh** 🤕. My bad. 

In the Contacts App, we can not have a `trashed: false` but you right, we can be more flexible with a better selector: 
```
"$or": [
         {
            "trashed": {
               "$exists": false
            }
         },
         {
            "trashed": false
         }
      ]
```

src/targets/browser/index.jsx
JF-Cozy: I think we want to import `./styles/index.styl` here, not `./styles.styl`

Crash--: So let's import ./styles/index.styl. It's better to be explicit with the extension

src/components/Selection/selection.spec.jsx
```diff
@@ -3,14 +3,18 @@ import PropTypes from 'prop-types'
...
 
+const DummyComponent = ({ title = '' }) => <span title={title} />
 DummyComponent.propTypes = {
   title: PropTypes.string
 }
+const DummyComponentWithSelection = selectionContainer(DummyComponent)
+
+// Uses a different version of react-redux
+// to prevent Enzyme's incompatibility with actual react-redux version
+jest.mock('react-redux', () => require('react-redux-test'))
```
Crash--, line 17: Can you link to an issue or something?

src/components/Modals/ContactCardModal.jsx
```diff
@@ -1,96 +1,85 @@
...
-            )
-          }}
-        </Query>
-      </Modal>
-    )
-  }
+  const resultContactById = useQuery(
+    queryContactById(id).definition,
+    queryContactById(id).options
+  )
```
Crash--, line 49: Making a function call is more expensive than reading a var, we should store the result somewhere. 

const queryContactById = buildContactQuery(id)

src/components/Modals/ContactCardModal.jsx
```diff
@@ -1,96 +1,85 @@
...
+
+  const showResult =
+    (!isCollectionLoading(resultContactById) ||
+      isCollectionInCache(resultContactById)) &&
+    (!isCollectionLoading(resultAllGroups) ||
+      isCollectionInCache(resultAllGroups))
+
+  const contact =
+    resultContactById.data !== null ? resultContactById.data[0] : {}
+  const allGroups = resultAllGroups.data !== null ? resultAllGroups.data : []
```
Crash--, line 63: not sure we need this `resultAllGroups.data !== null` anymore since we use `allGroups` only when `showResult` is true === data loaded, no ? Same for `resultContactById` ?

src/components/Modals/ContactCardModal.jsx
```diff
@@ -1,96 +1,85 @@
...
+  const resultContactById = useQuery(
+    queryContactById(id).definition,
+    queryContactById(id).options
+  )
+  const resultAllGroups = useQuery(
+    queryAllGroups.definition,
+    queryAllGroups.options
+  )
+
+  const showResult =
```
Crash--, line 55: nit: I don't like having this kind of naming. It doesn't tell us anything. 

const dataHaveBeenLoaded or something similar is way better IMHO.

src/helpers/queries.js
```diff
@@ -0,0 +1,34 @@
+import { Q, fetchPolicies } from 'cozy-client'
+import { DOCTYPE_CONTACTS } from './doctypes'
+
+export const cacheTime = 30000
```
Crash--, line 4: const olderThan30s = fetchPolicies.olderThan(30000)
...

fetchPolicy: olderThan30s

1/ more readable 
2/ only one assignation (not twice in every render)

src/helpers/utils.js
```diff
@@ -0,0 +1,17 @@
...
+  return col.fetchStatus === 'loading' || col.fetchStatus === 'pending'
+}
+
+export const isCollectionInCache = col => {
+  if (!col) {
+    console.warn('isCollectionInCache called on falsy value.') // eslint-disable-line no-console
+    return false
+  }
+  return col.lastFetch + cacheTime >= Date.now()
+}
```
Crash--, line 17: Why are you doing that? You don't need to check that. CozyClient and its fetch policy will do that for you. In our case, we just need to check if `lastFetch` exists to know if we have datas in the store. Nothing more / less.

Crash--: Also https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/utils.js#L30-L43 ;)

src/components/Modals/ContactCardModal.jsx
JF-Cozy: We can simplify by using `showResult` but at the first render the data are still null. But we can remove those const and use the data directly in the return part (wrap into `showResult`)

src/helpers/utils.js
Crash--: You pass a query definition to CozyClient. This QueryDefinition has an `as` and a fetchPolicy. CozyClient, will check if in its own store he has data for this `as`. If there is something, he will check if the `fetchPolicy` is respected or not. If it's `respected` then he doesn't `fetch` he will only read from its store, if not he will `fetch` the data https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/Query.jsx#L106-L117

src/components/Modals/ContactCardModal.jsx
Crash--: You right, the data will be null at the first render. But we don't use them at this time, no? We only read them if `showResult` (or `dataHaveBeenLoaded`) is true, right? So we don't need to deal with `null` data, right?

JF-Cozy: We only read them if showResult (or dataHaveBeenLoaded) is true, right? > Yes. Code updated according to `use the data directly in the return part`. What do you think ?

Crash--: Better! But I'm not very convinced by reading `resultContactById.data[0]`. Are you sure that getById() returns an array and not the object? 🤔

JF-Cozy:
> > (*)Add react-redux-test (react-redux not updated) to prevent broken tests and Enzyme incompatibility
> 
> Can you explain what was wrong with the latest react-redux? Is there any existing issue on Enzyme / Redux repo?

Yes, first message updated

src/components/Modals/ContactCardModal.jsx
JF-Cozy: Yes... It will not work otherwise... Besides, browser returns this
```
Failed prop type: Invalid prop `contact` of type `array` supplied to `ContactCard`, expected `object`.
```

src/components/Selection/selection.spec.jsx
Crash--: What if you use mount instead of shallow? 

I really don't like having 2 versions of react-redux, the test doesn't test the code used in our app...

src/components/Modals/ContactCardModal.jsx
Crash--: I think that Q().getById() should returns the related document, not an array containing one document. It should behave like the DocumentCollection.get(), no? If yes, we should open an issue on cozy-client

src/components/Selection/selection.spec.jsx
Crash--: Even if the hack is scoped, the test is not relevant anymore. We should create an issue on Contacts to rewrite this test with react-testing-lib. Since one of the next task will be to work on list's filter, maybe we will have to work on the selection part and fix this issue

Crash-- (approved):

src/components/Selection/selection.spec.jsx
JF-Cozy: issue : https://github.com/cozy/cozy-contacts/issues/665

src/components/Modals/ContactCardModal.jsx
JF-Cozy: issue : https://github.com/cozy/cozy-client/issues/764

PR #662, opened 2020-08-20
refactor: App to functional component

JF-Cozy:
This PR is a part [2] of multiple PR about the new sort logic in the application. Tests will maybe failed and probably don't after merging previous PR and rebasing this one.

- Refactor App to use functional component / hooks
- (*) Extract the main query to new file
- Cleanup browser index (app init) and store configuration

(*) the Query is still not using hooks. It is the very first refactor, the hook will comes in an other PR

### DIFF
--- a/src/components/ContentWrapper.jsx
+++ b/src/components/ContentWrapper.jsx
@@ -1,0 +1,51 @@
+import React from 'react'
+import { DOCTYPE_CONTACTS } from '../helpers/doctypes'
+import ContactsListDataLoader from './ContactsList/ContactsListDataLoader'
+import Header from './Header'
+import Toolbar from './Toolbar'
+import { Content } from 'cozy-ui/transpiled/react/Layout'
+import { Query } from 'cozy-client'
+
+const ContentWrapper = () => (
+  <Query
+    query={client =>
+      client
+        .all(DOCTYPE_CONTACTS)
+        .include(['accounts'])
+        .where({
+          $or: [
+            {
+              trashed: {
+                $exists: false
+              }
+            },
+            {
+              trashed: false
+            }
+          ],
+          _id: {
+            $gt: null
+          }
+        })
+        .indexFields(['_id'])
+    }
+  >
+    {({ data: contacts, fetchStatus, hasMore, fetchMore }) => {
+      return (
+        <>
+          {contacts.length >= 1 && <Header right={<Toolbar />} />}
+          <Content>
+            <ContactsListDataLoader
+              contacts={contacts}
+              fetchStatus={fetchStatus}
+              hasMore={hasMore}
+              fetchMore={fetchMore}
+            />
+          </Content>
+        </>
+      )
+    }}
+  </Query>
+)
+
+export default ContentWrapper
--- a/src/store/configureStore.js
+++ b/src/store/configureStore.js
@@ -8,7 +8,6 @@ import {
   getTracker,
   createTrackerMiddleware
 } from 'cozy-ui/transpiled/react/helpers/tracker'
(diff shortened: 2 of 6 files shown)


src/components/ContentWrapper.jsx
```diff
@@ -0,0 +1,46 @@
...
+          $exists: false
+        },
+        _id: {
+          $gt: null
+        }
+      })
+      .indexFields(['_id'])
+
+  return (
+    <Query query={query(client)}>
```
Crash--, line 26: You can simply <Query query={query}> and `client` will be automatically injected. No need to get it through useClient for instance

src/components/ContentWrapper.jsx
```diff
@@ -0,0 +1,44 @@
...
+const ContentWrapper = () => (
+  <Query
+    query={client =>
+      client
+        .all(DOCTYPE_CONTACTS)
+        .include(['accounts'])
+        .where({
+          trashed: {
+            $exists: false
+          },
```
Crash--, line 18: you can change the selector here too

src/store/configureStore.js
```diff
@@ -8,40 +8,35 @@ import {
...
-    middlewares.push(getReporterMiddleware(cozyClient))
-  } */
 
   const store = createStore(
-    reducers,
+    combineReducers({
+      appReducers,
+      cozy: client.reducer()
+    }),
+    persistedState,
```
Crash--, line 39: why `persistedState` is not in `combineReducers` anymore? We're changing the behavior by doing that. It should not be tagged as refactor if this is the case

src/targets/browser/index.jsx
```diff
@@ -1,121 +1,59 @@
...
+  const client = getClient()
   const persistedState = {}
   const store = configureStore(
     client,
     polyglot.t.bind(polyglot),
     persistedState
   )
+
   /** I don't know why I need to for this... But if I don't it seems that
    * we have a race between configureStore and initCozyBar resulting in
```
Crash--, line 34: initCozyBar doesn't exist anymore

Crash--:
Tests are red. 

Also, if I had to be a pain in the ass I'd say you should have made a commit for the hook translate, one for the breakpoint etc.

JF-Cozy:
> Tests are red.

Yes I know, as I said in the PR description `Tests will maybe failed and probably don't after merging previous PR and rebasing this one.`

> Also, if I had to be a pain in the ass I'd say you should have made a commit for the hook translate, one for the breakpoint etc.

Hmmm hooks like those just add a `const` and remove entries in the `export`. And a wrapper in the `breakpoint` case... so just few changes, but ok next time I will do it by hooks.

src/store/configureStore.js
JF-Cozy: Damned my bad, it's a copy/past from Banks

Crash--:
> Hmmm hooks like those just add a const and remove entries in the export. And a wrapper in the breakpoint case... so just few changes, but ok next time I will do it by hooks.

It's just way easier to review commit by commit with one commit dedicated to one thing.

JF-Cozy:
Can be merged, previous PR https://github.com/cozy/cozy-contacts/pull/661 was merged

JF-Cozy:
> > Tests are red.
> 
> Yes I know, as I said in the PR description `Tests will maybe failed and probably don't after merging previous PR and rebasing this one.`. So now previous PR is merged, this one rebased and it's ok.
> 

My mistake it's not in this PR :( I add it right now, so my apologize

Crash-- (approved):

PR #663, opened 2020-08-20
Use displayName in list's first column and don't sort contacts in list

JF-Cozy:
This PR is a part [3] of multiple PR about the new sort logic in the application. Tests will maybe failed and probably don't after merging previous PR and rebasing this one.

- Use the `displayName` of a contact to show contact's information in first column
- Refactor `ContactIdentity` to use `getInitials()` and `getFullname()` from cozy-client
- Refactor `ContactList` and  change it to don't sort contacts (they will be sorted directly in the App and then past to the ContactList in an other PR)

**Must be merged after https://github.com/cozy/cozy-contacts/pull/662**


JF-Cozy:
https://github.com/cozy/cozy-contacts/pull/662 merged, if nothing wrong this is a green light for merging this one

src/components/ContactCard/ContactIdentity.spec.jsx
```diff
@@ -47,6 +47,30 @@ describe('ContactIdentity', () => {
...
+    const app = mount(
+      <AppLike>
+        <ContactIdentity contact={contactWithoutName} groups={groupsMock} />
+      </AppLike>
+    )
+    const avatar = app.find('Avatar')
+    expect(avatar.prop('text')).toEqual('S')
+  })
+
+  it('should have the right informations even if the contact does not have a name and no email', () => {
```
Crash--, line 59:
```suggestion
  it('should have the right informations even if the contact does not have a name or email', () => {
```

src/components/ContactsList/ContactRow.jsx
```diff
@@ -20,35 +21,24 @@ class ContactRow extends Component {
...
-ContactRow.propTypes = {
-  contact: PropTypes.shape({
-    name: contactPropTypes.name,
-    email: PropTypes.arrayOf(contactPropTypes.email),
-    phone: PropTypes.arrayOf(contactPropTypes.phone),
-    metadata: contactPropTypes.metadata
-  }).isRequired,
 
+ContactRow.propTypes = {
+  contact: PropTypes.object.isRequired,
```
Crash--, line 41: why did you remove the shape?

src/components/ContactsList/Contacts/ContactIdentity.jsx
```diff
@@ -3,25 +3,28 @@ import PropTypes from 'prop-types'
...
+      <Avatar text={getInitials(contact)} size="small" />
+      <ContactName displayName={displayName} lastname={name.familyName} />
+      {isMyself && <MyselfMarker />}
+    </div>
+  )
 }
-ContactIdentity.defaultProps = {
-  myself: false
+ContactIdentity.propTypes = {
+  contact: PropTypes.object.isRequired
```
Crash--, line 27: you can use the shape

src/components/ContactsList/Contacts/ContactName.jsx
```diff
@@ -1,21 +1,28 @@
...
-    <span className="contact-lastname">{lastname}</span>
-  </div>
-)
+const ContactName = ({ displayName, lastname }) => {
+  const namesToDisplay = (displayName && displayName.split(' ')) || []
+  return (
+    <div className="u-ellipsis">
+      {namesToDisplay.map((name, key) => (
+        <span key={`display-${key}`}>
+          <span className={name === lastname ? 'contact-lastname' : 'u-ml-1'}>
```
Crash--, line 10: Can you share a screenshot of the list with : one contact having a lastname and a name, and one only with something to display but not a lastname ? I think the one without a lastname will start with a left margin and this is not what we want, right? 

Also, I think we can remove the contact-lastname class and use u-bold or something ;)

src/components/ContactsList/Contacts/ContactName.jsx
```diff
@@ -1,21 +1,28 @@
...
+            {name}
+          </span>
+          &nbsp;
```
Crash--, line 13: why this? don't the `u-ml-1` is there to handle the space between names?

src/components/ContactsList/__snapshots__/ContactsListDataLoader.spec.jsx.snap
```diff
@@ -41,33 +41,30 @@ exports[`ContactsListOrSpinner should display the contact List 1`] = `
...
+                  }
+                }
               >
                 
-                
-                <svg
-                  className="styles__icon___23x3R"
-                  height="16"
-                  style={Object {}}
-                  width="16"
```
Crash--, line 51: why this removal?

src/helpers/contactList.spec.js
```diff
@@ -0,0 +1,39 @@
...
+      B: [
+        { indexes: { byFamilyNameGivenNameEmailCozyUrl: 'B' }, name: 'Bernard' }
+      ],
+      C: [
+        { indexes: { byFamilyNameGivenNameEmailCozyUrl: 'C' }, name: 'Cleo' }
+      ],
+      EMPTY: [
+        { indexes: { byFamilyNameGivenNameEmailCozyUrl: '' }, name: '' },
+        { indexes: { byFamilyNameGivenNameEmailCozyUrl: {} }, name: {} }
+      ],
```
Crash--, line 30: weird to have this `empty` in the middle of the categories. Shouldn't we put in at the end?

src/components/ContactsList/Contacts/ContactName.jsx
JF-Cozy: the space between names should be unbreakable. Besides the margin is only used to make space between avatar and name.

src/helpers/contactList.spec.js
JF-Cozy: No I don't think so, the contactList doesn't sort anymore, but categorize only.

src/components/ContactsList/Contacts/ContactName.jsx
Crash--: >Besides the margin is only used to make space between avatar and name. 

OK but if the lastname is the first name to be displayed, we'll not have this margin. 

The margin should not be attached to this node, but its parent ?

src/helpers/contactList.spec.js
Crash--: So merging this PR will make the contacts list unordered? If yes, I think even for a dev release this is not wanted?

src/components/ContactsList/ContactRow.jsx
JF-Cozy: Because used like this on other places, so I picked the wrong example.
Now they are all fixed and using shape

src/components/ContactsList/__snapshots__/ContactsListDataLoader.spec.jsx.snap
JF-Cozy: Because for those scenarios (no name but email address or cozy url) there is no longer default avatar but avatar with initials

src/components/ContactsList/Contacts/ContactName.jsx
JF-Cozy: Yes you're right, fixed !

src/components/ContactsList/Contacts/ContactName.jsx
JF-Cozy: Margin are now fixed.
PR on cozy-ui incoming.

src/components/ContactsList/Contacts/ContactIdentity.jsx
```diff
@@ -1,27 +1,32 @@
...
+
+const ContactIdentity = ({ contact }) => {
+  const name = contact.name || {}
+  const displayName = getDisplayName(contact) || undefined
+  const isMyself = contact.metadata ? !!contact.metadata.me : false
+
+  return (
+    <div className="contact-identity">
+      <Avatar text={getInitials(contact)} size="small" />
+      <ContactName displayName={displayName} lastname={name.familyName} />
```
Crash--, line 23: We should avoid to change the name of a prop, we should keep familyName, no? 

Also, I'm wondering if `ContactName` should not receive the full `contact` object and call displayName() etc by itself.

JF-Cozy: Ok for `familyName`.

For `ContactName` I think it depends if we want to have a dumb component or not. For now the only treatment in it is done for styling, not to treat datas.

Crash--: I don't know, to me if the component is named `ContactX` it should handle a contact object. But it's a nit ;)

src/helpers/contactList.spec.js
Crash--: @JF-Cozy ?

src/components/ContactsList/Contacts/ContactName.jsx
```diff
@@ -1,21 +1,30 @@
...
-    &nbsp;
-    <span className="contact-lastname">{lastname}</span>
-  </div>
-)
+const ContactName = ({ displayName, lastname }) => {
+  const namesToDisplay = (displayName && displayName.split(' ')) || []
+  return (
+    <div className="u-ellipsis u-ml-1">
+      {namesToDisplay.map((name, key) => (
+        <span key={`display-${key}`}>
```
y-lohse, line 9: Do we need the extra `span`? Would something like `<span className={{'u-bold': name === lastname}}>{name}&nbsp;</span>` work?

src/helpers/contactList.spec.js
JF-Cozy: Yes it is, it's sorted based on `_id`.

For me it's a PR among others, which are all part of a feature. As long as the last PR of the feature is not merged, there is not the expected final behavior. But I understand that it's tricky to have an unexpected behavior even if it's temporary on a Master. I will correct it and use the part of this commit on the other PR.

Crash--: If you think you need to create several PR for one feature, you can:
- Create a feature branch like `newSort`
- Creates PR on this new branch 
- Once each PR are merged in this `newSort` branch, create a PR of this branch to master. 

We don't have a development branch, we should be able to deploy what we have on master at any time

JF-Cozy: Done :)

src/components/ContactsList/Contacts/ContactName.jsx
```diff
@@ -1,21 +1,30 @@
...
+import cx from 'classnames'
+
+const ContactName = ({ displayName, familyName }) => {
+  const namesToDisplay = (displayName && displayName.split(' ')) || []
+  return (
+    <div className="u-ellipsis u-ml-1">
+      {namesToDisplay.map((name, key) => (
+        <span
+          key={`display-${key}`}
+          className={cx({ 'u-fw-bold': name === familyName })}
```
Crash--, line 12: 👍

Crash-- (approved):

PR #664, opened 2020-08-20
Add fixtures to get contacts without names and thousands contacts

JF-Cozy:
- Add some flexibility with fixtures : we can now inject more than 2000 contacts in one shot and contacts without name


Crash-- (approved):
Nice 👍

PR #667, opened 2020-08-24
fix: Update cozy-ui and update/fix cozy-bar init

JF-Cozy:
Use last version of cozy-ui and cozy-bar and fix cozy-bar init


Crash-- (changes requested):
We can't rely on the import of the cozy-bar. 

In production, cozy-bar is served by the stack and injected in our application. So you can't import it like that. 

cozy-bar in dev mode is injected by cozy-script and we should not change that for now

JF-Cozy:
> We can't rely on the import of the cozy-bar.
> 
> In production, cozy-bar is served by the stack and injected in our application. So you can't import it like that.
> 
> cozy-bar in dev mode is injected by cozy-script and we should not change that for now

Ok :)
So I misunderstood how it works in Banks https://github.com/cozy/cozy-banks/blob/master/src/main.jsx#L31 . What's the difference I missed their ?

Crash--:
Good question. I didn't see this modification on Banks https://github.com/cozy/cozy-banks/pull/1824/files. @ptbrowne can you explain how it works for a web production build of banks?

ptbrowne:
It works differently as other apps and uses a bar that is bundled.

I wanted to test if it was really necessary to have the bar injected by the stack. At this moment, I think banks was always at the forefront of the bar versioning and we did not want to wait for the stack to upgrade its version.

I'll revert it back to the normal way of doing things, since it is confusing.

PR #668, opened 2020-08-24
Use two differents queries depends if index service was launched

JF-Cozy:
If service `keepIndexFullNameAndDisplayNameUpToDate` was already launched we want to sort contacts in a specific way. So we have to test that first, then query contacts in a way or in an other.

- Check if service was launched
- Fetches and sort contacts in two differents ways
- Refactor of fetch function to get the service

### DIFF
--- a/manifest.webapp
+++ b/manifest.webapp
@@ -57,7 +57,7 @@
     "triggers": {
       "description": "Used to sort contacts",
       "type": "io.cozy.triggers",
-      "verbs": ["GET"]
+      "verbs": ["GET", "POST"]
     }
   },
   "intents": [
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "cozy-doctypes": "1.72.2",
     "cozy-flags": "1.10.0",
     "cozy-logger": "1.6.0",
-    "cozy-ui": "35.39.0",
+    "cozy-ui": "35.40.1",
     "cozy-vcard": "0.2.18",
     "final-form": "4.10.0",
     "final-form-arrays": "1.1.0",
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,8 +1,9 @@
 /* global cozy */
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useCallback } from 'react'
 import { PropTypes } from 'prop-types'
 import flow from 'lodash/flow'
 
+import { useClient } from 'cozy-client'
 import { Main, Layout } from 'cozy-ui/transpiled/react/Layout'
 import Alerter from 'cozy-ui/transpiled/react/Alerter'
 import { Title } from 'cozy-ui/transpiled/react/Text'
@@ -10,13 +11,16 @@ import { useI18n, useBreakpoints } from 'cozy-ui/transpiled/react'
 import { Sprite as IconSprite } from 'cozy-ui/transpiled/react/Icon'
 import flag, { FlagSwitcher } from 'cozy-flags'
 import 'cozy-ui/transpiled/react/stylesheet.css'
+import log from 'cozy-logger'
 
 import withContactsMutations from '../connections/allContacts'
 import ContactsSelectionBar from './layout/ContactsSelectionBar'
 import { ModalManager } from '../helpers/modalManager'
-import { initFlags } from '../helpers/flags'
+import useFlags from './Hooks/useFlags'
 import container from './AppContainer'
 import ContentWrapper from './ContentWrapper'
+import SpinnerContact from './Components/Spinner'
+import { fetchNormalizedServiceByName } from '../helpers/fetches'
 
 const ContactsApp = props => {
   // HACK to avoid CozyBar error :
@@ -24,20 +28,54 @@ const ContactsApp = props => {
   // TODO : TO BE REMOVED
   const [cozyBarHack, setcozyBarHack] = useState(false)
 
-  useEffect(() => {
-    initFlags()
-    props.cleanTrashedGroups()
+  const [serviceToLaunch, setServiceToLaunch] = useState(null)
+  const [hasServiceBeenLaunched, setHasServiceBeenLaunched] = useState(null)
 
+  useFlags()
+  const client = useClient()
+  const { t } = useI18n()
+  const { isMobile } = useBreakpoints()
+  const { BarCenter } = cozy.bar
+  const { deleteContact, cleanTrashedGroups } = props
+
+  const setStateOfServiceToLaunch = useCallback(async () => {
+    const serviceToLaunch = await fetchNormalizedServiceByName(
+      client,
+      'keepIndexFullNameAndDisplayNameUpToDate'
+    )
+    setServiceToLaunch(serviceToLaunch)
+    setHasServiceBeenLaunched(
+      serviceToLaunch.current_state.last_success.length > 0
+    )
+  }, [])
+
+  useEffect(() => {
+    cleanTrashedGroups()
+  }, [])
+
+  useEffect(
+    () => {
+      setStateOfServiceToLaunch()
+    },
+    [setStateOfServiceToLaunch]
+  )
+
+  useEffect(() => {
     // HACK to be removed
     setTimeout(() => {
       setcozyBarHack(true)
     }, 0)
   }, [])
 
-  const { t } = useI18n()
-  const { isMobile } = useBreakpoints()
-  const { BarCenter } = cozy.bar
-  const { deleteContact } = props
+  // start keepIndexFullNameAndDisplayNameUpToDate service
+  // if never launched before
+  if (serviceToLaunch && hasServiceBeenLaunched === false) {
+    log(
+      'info',
+      `Executing keepIndexFullNameAndDisplayNameUpToDate service by Contacts app`
+    )
+    client.collection('io.cozy.triggers').launch(serviceToLaunch)
+  }
 
   return (
     <Layout monocolumn="true">
@@ -52,7 +90,11 @@ const ContactsApp = props => {
       <Main>
         {flag('switcher') && <FlagSwitcher />}
         <ContactsSelectionBar trashAction={deleteContact} />
-        <ContentWrapper />
+        {hasServiceBeenLaunched === null ? (
+          <SpinnerContact size="xxlarge" loadingType="fetching_contacts" />
+        ) : (
+          <ContentWrapper hasServiceBeenLaunched={hasServiceBeenLaunched} />
+        )}
         <Alerter t={t} />
         <ModalManager />
       </Main>
--- a/src/components/ContentWrapper.jsx
+++ b/src/components/ContentWrapper.jsx
@@ -1,51 +1,35 @@
 import React from 'react'
-import { DOCTYPE_CONTACTS } from '../helpers/doctypes'
-import ContactsListDataLoader from './ContactsList/ContactsListDataLoader'
+import { PropTypes } from 'prop-types'
+
+import { Content } from 'cozy-ui/transpiled/react/Layout'
+
 import Header from './Header'
 import Toolbar from './Toolbar'
-import { Content } from 'cozy-ui/transpiled/react/Layout'
-import { Query } from 'cozy-client'
+import ContactsList from './ContactsList/ContactsList.jsx'
+import SpinnerContact from './Components/Spinner'
+import fetchSortedContacts from './Hooks/fetchSortedContacts'
 
-const ContentWrapper = () => (
-  <Query
-    query={client =>
-      client
-        .all(DOCTYPE_CONTACTS)
-        .include(['accounts'])
-        .where({
-          $or: [
-            {
-              trashed: {
-                $exists: false
-              }
-            },
-            {
-              trashed: false
-            }
-          ],
-          _id: {
-            $gt: null
-          }
-        })
-        .indexFields(['_id'])
-    }
-  >
-    {({ data: contacts, fetchStatus, hasMore, fetchMore }) => {
-      return (
-        <>
-          {contacts.length >= 1 && <Header right={<Toolbar />} />}
-          <Content>
-            <ContactsListDataLoader
-              contacts={contacts}
-              fetchStatus={fetchStatus}
-              hasMore={hasMore}
-              fetchMore={fetchMore}
-            />
-          </Content>
-        </>
-      )
-    }}
-  </Query>
-)
+const ContentWrapper = ({ hasServiceBeenLaunched }) => {
+  const [hasContactsToShow, contacts] = fetchSortedContacts(
+    hasServiceBeenLaunched
+  )
+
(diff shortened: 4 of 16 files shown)


src/helpers/contentWrapper.js
```diff
@@ -0,0 +1,67 @@
...
+    resultWithIndexes.fetchStatus === 'loaded' &&
+    resultWithoutIndexes.fetchStatus === 'loaded'
+      ? 'loaded'
+      : 'loading'
+
+  const hasMore = resultWithIndexes.hasMore || resultWithoutIndexes.hasMore
+
+  const fetchMore = resultWithIndexes.hasMore
+    ? resultWithIndexes.fetchMore
+    : resultWithoutIndexes.fetchMore
```
Crash--, line 37: What happens if both have fetchMore?

src/helpers/contentWrapper.js
```diff
@@ -0,0 +1,67 @@
...
+  const hasData = result.data && result.data.length > 0
+
+  if (!isInError && hasData) {
+    const contactsWithIndex = result.data.map(contact =>
+      updateIndexFullNameAndDisplayName(contact)
+    )
+    data = sortBy(contactsWithIndex, [
+      'indexes.byFamilyNameGivenNameEmailCozyUrl'
+    ])
+  }
```
Crash--, line 65: Are we doing this `updateIndexFullNameAndDisplayName` even after a fetchMore() is called? 🤔

src/helpers/contentWrapper.js
JF-Cozy: Both should have fetchMore anyway, at least it will be a function doing nothing.
As long as `hasMore` isn't at `false`, it must fetch anyway. If it's the first request that has `hasMore`, it's its `fetchMore` that's passed, otherwise it's the `fetchMore` of the second request.

src/helpers/contentWrapper.js
JF-Cozy: Oh yes good catch, adding `&& !hasMore`

package.json
```diff
@@ -66,7 +66,7 @@
     "cozy-doctypes": "1.72.2",
     "cozy-flags": "1.10.0",
     "cozy-logger": "1.6.0",
-    "cozy-ui": "35.39.0",
```
Crash--, line 69: This commit should not be `chore` since `chore` is reserved for build tool or something that is not published in production. Since cozy-ui is included in the production build, `chore` is not the right tag. 

Either `fix` if this upgrade brings a fix for Contacts, or `feat` if it brings a new feature or somethign else

src/helpers/contentWrapper.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Get contacts with indexes,
+ * update contacts without indexes and
+ * sort them by indexes
+ * @param {array} contactsWithIndexes - contacts with indexes
+ * @param {array} contactsWithNoIndexes - contacts without indexes
+ * @returns {array} contacts with indexes and sorted
+ */
+export const updateAndSortByFamilyNameGivenNameEmailCozyUrl = (
```
y-lohse, line 14: Nit: the `update` in the name led me to think the function would perform server-side updates. IMO something like `getContactsSortedBy[etc]` might be clearer.

src/components/ContentWrapper.jsx
```diff
@@ -1,51 +1,80 @@
...
+    false
+  )
+  const [sortedContacts, setSortedContacts] = useState([])
+
+  const fetchesAreFinished =
+    isContactsWithIndexesFinished && isContactsWithNoIndexesFinished
+
+  // start keepIndexFullNameAndDisplayNameUpToDate service
+  // if never launched before
+  if (serviceToLaunch && isServiceLaunched === false) {
```
y-lohse, line 41: Why are we launching the service in this component and not in App?

src/components/ContentWrapper.jsx
```diff
@@ -1,51 +1,80 @@
...
+    )
+    client.collection('io.cozy.triggers').launch(serviceToLaunch)
+  }
+
+  if (!fetchesAreFinished) {
+    return <SpinnerContact size="xxlarge" loadingType="fetching_contacts" />
+  }
+
+  if (fetchesAreFinished && !hasAlreadySortedContacts) {
+    const reworkedContacts = isServiceLaunched
```
y-lohse, line 54: Reading this line I realise `isServiceFinished` or `hasServiceBeenLaunched` might be a more suitable name :)

src/components/ContentWrapper.jsx
```diff
@@ -1,51 +1,80 @@
...
+      ? contactsWithIndexes.concat(contactsWithNoIndexes)
```
y-lohse, line 55: I'm confused about why this works: if the service has run, do we still have `contactsWithNoIndexes`? And if we do, don't they need to be re-sorted?

y-lohse (commented):
Since I took a break, not everyhting is super clear to m so I mostly have suggestions for name improvements 😁

manifest.webapp
```diff
@@ -57,7 +57,7 @@
     "triggers": {
       "description": "Used to sort contacts",
       "type": "io.cozy.triggers",
-      "verbs": ["GET"]
+      "verbs": ["GET", "POST"]
```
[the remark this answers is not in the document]
JF-Cozy: For `client.collection('io.cozy.triggers').launch(serviceToLaunch)` in `ContentWrapper`.
`.launch()` needs `POST` permission

src/components/ContentWrapper.jsx
JF-Cozy: Oui on peut avoir encore des `contactsWithNoIndexes` car même si le service a déjà été lancé une fois, on peut par la suite créer un nouveau contact sans index (via un connecteur par exemple).

src/helpers/contentWrapper.js
JF-Cozy: Suite à discussion récente, `get` quant à lui fait penser à une synchronisation.

Que penses-tu de `reworkAndSortByFamilyNameGivenNameEmailCozyUrl` ?

src/helpers/contentWrapper.js
```diff
@@ -0,0 +1,53 @@
+import { useState } from 'react'
```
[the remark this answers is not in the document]
JF-Cozy: Ok `updateAndSortByFamilyNameGivenNameEmailCozyUrl` moved in contacts helper, and ` useFetchContacts` in `./src/components/Hooks`

src/helpers/contentWrapper.js
y-lohse: C'est pas mal, sinon `consolidateAnd…`? Ou `harmonizeAnd…`? Mais rework me va aussi, un truc dans le genre quoi 👍

src/components/ContentWrapper.jsx
y-lohse: Je comprend, mais dans ce cas les contacts sans index ne seront pas triés et seront à la fin de la liste non ? Pourquoi ne pas faire le tri côté client dans ce cas la aussi ?

src/components/ContentWrapper.jsx
y-lohse: @JF-Cozy ping Just in case you haven't seen this one, it's marked as outdated by github but it's still relevant :)

JF-Cozy: I had component reload issue before but it's no longer right. It is back in App.

src/components/ContentWrapper.jsx
JF-Cozy: Oui c'est ce comportement qui est attendu.
Parce qu'on veut éviter autant que possible de faire du traitement en front. Dans l'autre cas on n'a pas le choix, mais l'autre cas est voué à disparaître quand les contacts auront tous migré et que le service aura été lancé au moins une fois sur toutes les instances. Pour plus de précision il y a les infos sur le paper https://paper.dropbox.com/doc/SEC_0012_CONTACTS_Beta--A6YS4mP2ytOsATqmPovCsJ~dAg-8TAfiAilZTySqvdaYePjQ

src/helpers/contentWrapper.js
JF-Cozy: Va pour `harmonizeAnd`

src/components/App.jsx
```diff
@@ -1,44 +1,78 @@
...
+      initFlags()
+      props.cleanTrashedGroups()
+      setStateOfServiceToLaunch()
+
+      // HACK to be removed
+      setTimeout(() => {
+        setcozyBarHack(true)
+      }, 0)
+    },
+    [setStateOfServiceToLaunch]
```
Crash--, line 58: To me, `useEffect` should serve only one purpose, not several. Specially when several methods in this useEffect do not share the dependency. It the identity of `setStateOfServiceToLaunch` changes, the effect will fire again. 

By creating a custom hook for each purpose, it's easier to unit test them, and prevent unintended executions.

src/components/ContentWrapper.jsx
```diff
@@ -1,51 +1,65 @@
...
+  return (
+    <>
+      {sortedContacts.length >= 1 && <Header right={<Toolbar />} />}
+      <Content>
+        <ContactsList contacts={sortedContacts} />
+      </Content>
+    </>
+  )
+}
+
```
Crash--, line 59: I would like to see a few tests covering all this function. I'll be more confident on the behavior that I'm right now.

y-lohse: 👍  — and I wonder if it would make sense to extract the logic of this component into a standalone hook? Not only would it be easier to test, it would also mean the component only would only do something like `fetchSortedContacts`, and the implementation of `fetchSortedContacts` can change, or it can be easily reused elsewhere? Just a suggestion though.

src/components/App.jsx
```diff
@@ -1,43 +1,81 @@
...
 const ContactsApp = props => {
   // HACK to avoid CozyBar error :
   // you tried to use the CozyBar API (BarCenter) but the CozyBar is not initialised yet via cozy.bar.init
   // TODO : TO BE REMOVED
   const [cozyBarHack, setcozyBarHack] = useState(false)
 
+  const [serviceToLaunch, setServiceToLaunch] = useState(null)
+  const [hasServiceBeenLaunched, setHasServiceBeenLaunched] = useState(null)
+
+  useFlags()
```
y-lohse, line 34: I personally find it a bit weird when we have custom hooks that return nothing, but… it's not wrong 😁. No need to change it imo, but i wonder if that pattern is encouraged or not.

y-lohse (approved):
Thanks! We could use some tests on the fetch/sort logic as @Crash-- mentioned, but other  than that LGTM.

src/components/App.jsx
JF-Cozy: Oui j'ai eu la même réflexion de mon côté et j'ai hésité entre ça ou mettre un useEffect avec l'appel dedans... c'est un hook qui ne retourne rien mais je me suis dit qu'au moins il était autonome avec son useEffect embarqué.
Je peux rechanger sans problème si c'est choquant.

src/components/Hooks/fetchSortedContacts.js
```diff
@@ -0,0 +1,51 @@
...
+      : harmonizeAndSortByFamilyNameGivenNameEmailCozyUrl(
+          contactsWithIndexes,
+          contactsWithNoIndexes
+        )
+
+    setHasAlreadySortedContacts(true)
+    setSortedContacts(reworkedContacts)
+  }
+
+  return [true, sortedContacts]
```
y-lohse, line 48: If we return a static value here, why return it at all?

JF-Cozy: it returned at false in other condition L32 (if !fetchesAreFinished)

src/components/ContentWrapper.jsx
```diff
@@ -1,51 +1,35 @@
...
+import ContactsList from './ContactsList/ContactsList.jsx'
+import SpinnerContact from './Components/Spinner'
+import fetchSortedContacts from './Hooks/fetchSortedContacts'
+
+const ContentWrapper = ({ hasServiceBeenLaunched }) => {
+  const [hasContactsToShow, contacts] = fetchSortedContacts(
+    hasServiceBeenLaunched
+  )
+
+  if (!hasContactsToShow) {
```
Crash--, line 17: I don't like this wording since, if we have 0 contact, I expect to have "noContactsToShow" but I don't want to have a spinner.

src/components/Hooks/fetchSortedContacts.js
Crash--: We can return [fetchesAreFinished, sortedContacts] without any if / else, no?

src/components/Hooks/useFetchContacts.js
```diff
@@ -0,0 +1,31 @@
...
+  const [contacts, setContacts] = useState([])
+  const { data, fetchStatus, hasMore, fetchMore } = useQuery(
+    query.definition,
+    query.options
+  )
+
+  if (fetchStatus === 'loaded' && !hasMore) {
+    if (isFetchFinished) return [isFetchFinished, contacts]
+    setContacts(data)
+    setIsFetchFinished(true)
```
Crash--, line 21: Why do we need to have a `isFetchFinished` state? 


```
if (fetchStatus === 'loaded' && !hasMore) {
setContacts(data)
return [true, contacts]
}
```

Is not fine?

src/components/Hooks/fetchSortedContacts.js
JF-Cozy: Right, I'll keep only the first one `if (fetchesAreFinished && !hasAlreadySortedContacts)`

src/components/ContentWrapper.jsx
JF-Cozy: What about `isFetchFinished` ?

src/components/Hooks/useFetchContacts.js
JF-Cozy: No because `fetchStatus === 'loaded' && !hasMore`is the final state and don't change. So the condition is always true at the end. And `setContacts(data)` reload the component, so it will create an infinite loop.

But actually we can simplify more, thanks to you :)

src/helpers/contentWrapper.js
ptbrowne: When we transform the data to conform to a given schema or to make sure that is coherent with other documents, I like to use `normalize`. This is also what is used in the `cozy-client` codebase to make sure that data from the server is consistent between each call.

I would reserve "harmonize" for cases where the each item of the data needs to be tweaked a bit to work well with other items. Here, the normilization denotes that each datum is tweaked individually.

The fact that the function needs a "And" in its name is a smell IMHO. It indicates that the function is doing too much. It should be clearer if the caller does one function call to normalize and one function call to sort.

The name of the function here is a bit too telling also I think, the sort key is a detail of implementation that should not leak to the caller.

PR #675, opened 2020-09-01
fix: New sort

JF-Cozy:
- Groups button in contact card now show groups list correctly
- After editing a contact, the list now show the correct contact informations
- Remove useless Hooks

We drop Hooks and come back to `Query` component. Here some explanations : 

- We tried with `queryAll` but no re-render after editing contact.
- We tried with `useQuery` (unwanted re-render after clicking on a contact) or `useQueries` (using `useQuery` anyway).
By using two `useQuery` with `if hasMoreToFetch then fetchMore()`, after showing the list the application freeze 3-4sec. Besides, it takes 2-3sec to show the modal after clicking a contact (even though we move `onclick showModal` from `ContactsList` to `ContactRow` or a deeper component) and 1-2sec to remove it, because of `ContentWrapper` re-rendering. Tried with and without `fetchPolicy`.
- We tried with `Query` and it doesn't re-render after clicking on a contact, and doesn't re-render `ContentWrapper` (only itself) after editing a contact.


src/components/Hooks/fetchSortedContacts.js
```diff
@@ -21,30 +23,23 @@ const fetchSortedContacts = hasServiceBeenLaunched => {
...
+      if (fetchesAreFinished) {
+        const reworkedContacts = reworkContacts(
+          hasServiceBeenLaunched,
           contactsWithIndexes,
           contactsWithNoIndexes
         )
+        setSortedContacts(reworkedContacts)
+      }
+    },
+    [JSON.stringify(contactsWithIndexes), JSON.stringify(contactsWithNoIndexes)]
```
Crash--, line 40: Are you doing that because you want a deep equality check? https://github.com/kentcdodds/use-deep-compare-effect should be better. 

Even with the use of this deepCompareEffect, it's not confortable with this kind of check. No idea for now to how to do it better. 

Maybe @y-lohse or @ptbrowne have a few 💡

JF-Cozy: Yes it's because as it's said in use-deep-compare-effect `even if none of the properties changed, your effect will get called anyway` but also because React can't compare array or object unless they are stringified.

Crash--: https://medium.com/better-programming/why-you-shouldnt-use-json-stringify-to-compare-objects-in-javascript-c9a16b7331e 🎁

JF-Cozy: Damned !

y-lohse: I'm confused, why do we want to switch to a `useEffect`?

src/helpers/contacts.js
```diff
@@ -126,3 +126,27 @@ export const harmonizeAndSortByFamilyNameGivenNameEmailCozyUrl = (
...
+ * @param {array} contactsWithNoIndexes - Contacts without indexes
+ * @returns {array} Sorted and harmonized contacts
+ */
+export const reworkContacts = (
+  hasServiceBeenLaunched,
+  contactsWithIndexes,
+  contactsWithNoIndexes
+) => {
+  // hasServiceBeenLaunched will be deprecated when
+  // all instances have been migrated.
```
y-lohse, line 143: Are we actually planning a migration for this ? Or do we plan to wait until a certain date?

src/components/Hooks/fetchSortedContacts.js
JF-Cozy: @Crash--  For what I saw https://twitter.com/dan_abramov/status/1104414272753487872 

`useEffect` is used to simply do something (or not) depends on a data comparison. `setSortedContacts` is triggered only if queries result change, to prevent infinite loop

src/helpers/contacts.js
JF-Cozy: We plan to wait until a certain date

src/components/Hooks/fetchSortedContacts.js
JF-Cozy: @Crash-- I'm thinking about JS objects and stringify. Here it's an array of fetched data, do you think we'll encouter the same problem ?

src/helpers/contacts.js
Crash--: We do not plan anything yet. We can't wait until a certain date since we're not sure the app will be updated at some point. 

We have two options: 
- 1/ a rundeck action that performs this migration on all our cozies 
- 2/ Keep this code for ever. The code is small enough to keep it imho

src/components/Hooks/fetchSortedContacts.js
ptbrowne: > but also because React can't compare array, or object, or array of objects unless they are stringified.

As far as the data is coming from cozy-client (I see that it is coming from a useQuery call, thus it comes from the cozy-client store), it is safe to assume that it has been treated as an immutable object : inside cozy-client, we do not mutate objects in place, precisely so that you can rely on reference equality checks downstream.

To stringify the thousands of contacts to check if we need to render is very inefficient :/

I see two potential solutions:

- `useMemo` can be used here.
- Using reselect selectors so that there is a sorted "view" available for multiple objects to rely on.

src/helpers/contacts.js
JF-Cozy: By 'certain date' I meant the day all cozies instance was migrated... my bad it was confusing.

src/components/Hooks/fetchSortedContacts.js
y-lohse: Another idea, sounds too simple to work so I'm probably missing something: don't use `useEffect`, and don't store the `sortedContacts` in a `useState`, simply return them. 

As far as I understand, `useEffect` is only there to prevent the infinite loop triggered by `setSortedContacts`. But I don't see why we keep the sorted contacts around anyway, they are computed dynamically every time the hook runs..?

src/helpers/contacts.js
y-lohse: I just saw the comment was already there before, sorry for bringing it up again. That being said, if we don't have a migration plan, and no set date to remove this code, imo we should remove the comment as well.

src/components/Hooks/fetchSortedContacts.js
JF-Cozy: This will cause performance issue because everything is triggered/re-rendered after clicking on a contact in the list

Crash--: Why that? What is the root cause of this new render?

JF-Cozy: Hooks calls rerender the component not matter what. By default the useQuery hook make 3 renders to get data, even more if we need to fetchMore, as well as its "parents". So I think there is too much useless render.
I found https://github.com/facebook/react/issues/14476#issuecomment-471199055 and will explore `useMemo` way

ptbrowne: The 3 renders that you talk about are the initial renders, not why we need a full re-render of the list after a click on the contact.

ptbrowne: The full list is re-rendered here because the contact modal is shown at the list level. If it was done at row level, only the row would be re-rendered.

JF-Cozy: I tried but the behavior is the same : `ContentWrapper` and so the list, are re-rendered

src/components/ContentWrapper.jsx
```diff
@@ -28,6 +44,42 @@ const ContentWrapper = ({ hasServiceBeenLaunched }) => {
...
+                  contactsWithNoIndexesResult={contactsWithNoIndexesResult}
+                />
+              )
+            }}
+          </Query>
+        )
+      }}
+    </Query>
+  )
+}
```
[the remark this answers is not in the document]
JF-Cozy: There is already a spinner until both are finished. It's managed by `ContentWrapperResult`.

Yes we can show a spinner while the first one is not finished, and only then launch the second query. This will make no performance difference, but a visual one : when the first one is finished, there is again a spinner for the second one, with a little visual glitch (because the second one start at default position, not where the first one was before)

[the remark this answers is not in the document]
Crash--: > This will make no performance difference 

Why that? If the first query changes (let's say its fetchStatus), it will rerender its child, and then re-render que second query (and then make again the http call since there is no fetchPolicy on it). 

Here, if the first query changes and so re-render, all the children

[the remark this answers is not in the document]
Crash--: Oh right, Query makes the call in DidMount. https://github.com/cozy/cozy-client/blob/master/packages/cozy-client/src/Query.jsx#L116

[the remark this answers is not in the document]
JF-Cozy: There is 2 POST by queries : one for `io.cozy.contacts/_index` and one for `io.cozy.contacts/_find`

I ran a test and there is the results :

- With the actual code :
query1 : 3 renders
query2 : 6 renders
total of 3 couples of POST

Behavior :
render query1 then query2 then
trigger 2 couples of POST (1 initial one by queries) then
render query2 then query1 then query2 then
trigger 1 couple of POST (query1 fetchMore) then
render query2 then query2 then query1 then query2

- With spinner while query1 has not finished to fetch all datas
query1 : 3 renders
query2 : 2 renders
total of 3 couples of POST

Behavior :
render query1 then
trigger 1 couple of POST (initial one) then
render query1 then
trigger 1 couple of POST (query1 fetchMore) then
render query1 then query2 then
trigger 1 couple of POST (query2 initial one) then
render query2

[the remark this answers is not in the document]
JF-Cozy: I just tested it with 6400+ contacts and there is no performance difference when the list shows up.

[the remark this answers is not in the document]
JF-Cozy: What do you think ?

Crash-- (approved):

PR #676, opened 2020-09-01
fix: Error Invalid time value for service keepIndexFullNameAndDisplayNameUpToDate

JF-Cozy:
- Add some documentations for Fixtures and Services
- Fix bug "Invalid time value" when accessing current_state.last_success property of a trigger

When a service was never launched, current_state.last_success is undefined

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,20 @@ Cozy's apps use a standard set of _npm scripts_ to run common tasks, like watch,
 ### Fixtures
 
 A fixture file is available, you can import its data with :
+
 ```
 $ yarn fixtures
 ```
+
+You can also import more than 2000 contacts in one batch with this other command :
+
+```
+$ yarn fixtures:massive
+```
+
+### Services
+
+Services can be triggered (and tested manually) by running `yarn services:[ServiceName]`. You can take a look to the [cozy-konnector-dev documentation][cozy-konnector-dev] for more informations.
 
 ### Run it inside the VM
 
@@ -152,6 +163,7 @@ cozy-contacts is developed by cozy and distributed under the [AGPL v3 license][a
 [bill-doctype]: https://github.com/cozy/cozy-konnector-libs/blob/master/models/bill.js
 [konnector-doctype]: https://github.com/cozy/cozy-konnector-libs/blob/master/models/base_model.js
 [konnectors]: https://github.com/cozy/cozy-konnector-libs
+[cozy-konnector-dev]: https://github.com/konnectors/libs/tree/master/packages/cozy-jobs-cli#cozy-konnector-dev
 [agpl-3.0]: https://www.gnu.org/licenses/agpl-3.0.html
 [contribute]: CONTRIBUTING.md
 [tx]: https://www.transifex.com/cozy/
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -2,6 +2,7 @@
 import React, { useState, useEffect, useCallback } from 'react'
 import { PropTypes } from 'prop-types'
 import flow from 'lodash/flow'
+import get from 'lodash/get'
 
 import { useClient } from 'cozy-client'
 import { Main, Layout } from 'cozy-ui/transpiled/react/Layout'
@@ -46,7 +47,7 @@ const ContactsApp = props => {
       )
       setServiceToLaunch(serviceToLaunch)
       setHasServiceBeenLaunched(
-        serviceToLaunch.current_state.last_success.length > 0
+        get(serviceToLaunch, 'current_state.last_success', '').length > 0
       )
     },
     [client]
--- a/src/helpers/fetches.spec.js
+++ b/src/helpers/fetches.spec.js
@@ -1,8 +1,9 @@
+import CozyClient from 'cozy-client'
+
 import { fetchContactsToUpdate, fetchNormalizedServiceByName } from './fetches'
 import { updateIndexFullNameAndDisplayName } from './contacts'
 import { johnDoeContact } from './testData'
 
-import CozyClient from 'cozy-client'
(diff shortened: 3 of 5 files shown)


src/targets/services/keepIndexFullNameAndDisplayNameUpToDate.js
```diff
@@ -18,7 +19,11 @@ export const keepIndexFullNameAndDisplayNameUpToDate = async () => {
   )
   const contactsToUpdate = await fetchContactsToUpdate(
     client,
-    normalizedService.current_state.last_success
+    get(
+      normalizedService,
+      'current_state.last_success',
+      '0000-01-01T00:00:00.00Z'
```
JF-Cozy, line 25: Comme le service ne traite que les contacts plus récents qu'une certaine date, mettre la date à "0" permet de traiter tous les contacts (dans le cas ou le service n'a jamais été exécuté)

README.md
```diff
@@ -46,6 +46,18 @@ A fixture file is available, you can import its data with :
 $ yarn fixtures
 ```
 
+You can also import more than 2000 contacts in one batch with this other command :
+```
```
ptbrowne, line 50: Can you add a line break before \`\`\`, it ensures that the markdown parser from the docs generator understands correctly the code block (not the same parser as GitHub).

README.md
```diff
@@ -46,6 +46,18 @@ A fixture file is available, you can import its data with :
...
+$ yarn fixtures:massive
+```
+
+### Services
+
+Services can be triggered (and so tested) with `cozy-konnector-dev` by using yarn command. A token will be generated the first time to grant access to `cozy-konnector-dev`, and you may need to delete it (`.token.json`) and regenerate it again after a while.
+```
```
ptbrowne, line 57: Same here

src/targets/services/keepIndexFullNameAndDisplayNameUpToDate.js
ptbrowne: Nit: From the point of the reader, it is a bit hard by reading the function call to understand that the 3rd attribute of fetchContactsToUpdate is the minimum update date.

README.md
ptbrowne: I think this line belongs to a general "How to run a service" in the cozy-konnector-libs repo. The docs from contacts could link to the general docs. Otherwise we'd have to repeat this line in each one of our repo. What do you think ?

src/components/App.jsx
```diff
@@ -1,7 +1,7 @@
 /* global cozy */
 import React, { useState, useEffect, useCallback } from 'react'
 import { PropTypes } from 'prop-types'
-import flow from 'lodash/flow'
+import { flow, get } from 'lodash'
```
y-lohse, line 4: Nit: we prefer multiple lodash imports to keep bundle sizes small (https://nolanlawson.com/2018/03/20/smaller-lodash-bundles-with-webpack-and-babel/)

y-lohse (approved):
LGTM with the small changes mentioned :)

src/components/App.jsx
ptbrowne: In this case, I think the bundle will be small anyway since we are tree shaking. I think the perf gain would be more in dev mode.

y-lohse: Yep, presumably on build times as well since we don't need to parse the whole lodash bundle with specific imports.

README.md
JF-Cozy: See https://github.com/konnectors/libs/pull/707

README.md
```diff
@@ -42,10 +42,25 @@ Cozy's apps use a standard set of _npm scripts_ to run common tasks, like watch,
...
 
+You can also import more than 2000 contacts in one batch with this other command :
+
+```
+$ yarn fixtures:massive
+```
+
+### Services
+
+Services can be triggered (and so tested) by using yarn command (use `cozy-konnector-dev` [see the documentation here][cozy-konnector-dev])
```
Crash--, line 58:
```suggestion
Services can be triggered (and tested manually) by running `cozy-konnector-dev` [see the documentation here][cozy-konnector-dev])
```

src/helpers/fetches.spec.js
```diff
@@ -7,14 +7,15 @@ const client = new CozyClient({})
 
 describe('fetchContactsToUpdate', () => {
   it('should returns only contact to update', async () => {
-    const lastSuccess = new Date()
     const contactToUpdate = johnDoeContact
     const contactUpToDate = updateIndexFullNameAndDisplayName(johnDoeContact)
```
Crash--, line 11: Est-ce qu'on devrait pas tester le cas ou le contact a une date inférieure / supérieure à ce qu'on attend pour vérifier le comportement ?

JF-Cozy:
> LGTM with the small changes mentioned :)
> 
> edit: Although a test for the thing that used to b broken would be a good addition!

The problem here was passing an `undefined' argument. Instead of making sure the argument to pass is not `undefined', we can check the argument passed inside the function. Would that sound better ?

It doesn't crash with an `undefined` argument, it throws an error. I think it's a typical case for TypeScript here.

Did you have anything else in mind ?

y-lohse:
@JF-Cozy I think the fix is correct as it is, no need to change it.

> It doesn't crash with an undefined argument, it throws an error.

What i had in mind was to call the service with a `last_success` that is undefined in a test, and make sure it doesn't throw an error anymore.

src/helpers/fetches.spec.js
JF-Cozy: As we mock the query result and the date is used in the query definition, we can't test that

Crash-- (approved):

PR #679, opened 2020-09-06
fix: Rerender ContactRow only on contact or breakpoints change

JF-Cozy:
- Lines in the list are only re-rendered if there is a change of contact or breakpoints, which avoids re-rendering all lines when only one contact is changed.
- Remove useless `withModalContainer` (add props to show/hide modal) HOC on `ContactRow`

edit : will add some tests

### DIFF
--- a/src/components/ContactsList/ContactRow.jsx
+++ b/src/components/ContactsList/ContactRow.jsx
@@ -7,7 +7,8 @@ import { models } from 'cozy-client'
 import withBreakpoints from 'cozy-ui/transpiled/react/helpers/withBreakpoints'
 
 import { fullContactPropTypes } from '../ContactPropTypes'
-import withModalContainer from '../HOCs/withModal'
+import withModal from '../HOCs/withModal'
+import ContactCardModal from '../Modals/ContactCardModal'
 import ContactWithSelection from './ContactSelection'
 import ContactPhone from './Contacts/ContactPhone'
 import ContactIdentity from './Contacts/ContactIdentity'
@@ -49,7 +50,7 @@ class ContactRow extends Component {
   render() {
     const {
       contact,
-      onClick,
+      showModal,
       breakpoints: { isMobile }
     } = this.props
     const email = getPrimaryEmail(contact) || undefined
@@ -57,7 +58,14 @@ class ContactRow extends Component {
     const cozyUrl = getPrimaryCozy(contact) || undefined
 
     return (
-      <div className="contact" onClick={onClick}>
+      <div
+        className="contact"
+        onClick={() =>
+          showModal(
+            <ContactCardModal onClose={this.hideContactCard} id={contact._id} />
+          )
+        }
+      >
         <ContactWithSelection contact={contact} />
         <ContactIdentity contact={contact} />
         {!isMobile && <ContactEmail email={email} />}
@@ -70,11 +78,10 @@ class ContactRow extends Component {
 
 ContactRow.propTypes = {
   contact: fullContactPropTypes.isRequired,
-  onClick: PropTypes.func
+  showModal: PropTypes.func.isRequired
 }
 ContactRow.defaultProps = {
-  selection: null,
-  onClick: null
+  selection: null
 }
 
-export default withBreakpoints()(withModalContainer(ContactRow))
+export default withBreakpoints()(withModal(ContactRow))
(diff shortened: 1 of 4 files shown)


JF-Cozy:
@y-lohse I think there is no need to test more props, others (id, keys, onClick) are not relevant imo

src/components/ContactsList/ContactRow.jsx
```diff
@@ -15,6 +15,24 @@ import ContactEmail from './Contacts/ContactEmail'
 const { getPrimaryCozy, getPrimaryPhone, getPrimaryEmail } = models.contact
 
 class ContactRow extends Component {
+  shouldComponentUpdate(nextProps) {
+    const { contact, breakpoints } = this.props
+
+    const hasContactBeenUpdated =
+      contact._rev !== nextProps.contact._rev ||
+      get(contact, 'cozyMetadata.updatedAt', '') !==
+        get(nextProps, 'contact.cozyMetadata.updatedAt', '')
```
ptbrowne, line 24: * Why is the 3rd parameter '' , and not nothing ? Is there value in this ?
* I think this could be extracted in a `hasDocBeenUpdated` function, this could be helpful to test it, and next maybe to extract it in cozy-client (maybe not now).

ptbrowne (dismissed):

y-lohse (changes requested):
> I think there is no need to test more props, others (id, keys, onClick) are not relevant imo

`key` is not a real prop, so that one can be ignored. But `id` and `onClick` are totally relevant. The component is now ignoring changes to all props except 2, but there's no way to know that from outside the component. If we reuse it and pass a changing `onClick` prop for example, it will be ignored and it's going to be hard to figure out why.

It's also going to be a problem right now because we haven't changed the [onClick](https://github.com/cozy/cozy-contacts/blob/master/src/components/ContactsList/ContactsList.jsx#L59) prop to a stable reference. It might still work because we only use the `id` prop of the contact, but still.

src/components/ContactsList/ContactRow.jsx
JF-Cozy: > Why is the 3rd parameter '' , and not nothing ? Is there value in this ?
To return empty string if cozyMetadata.updatedAt is undefined... but will work with undefined value too

ptbrowne: I know what the 3rd argument of get does ;) If it works with `undefined`, let's not put the empty string. We should not have unmeaningful default values : here I wondered why the empty value of a Date field was changed to be an empty String, not null/undefined, and if you had encountered any weird bug.

src/components/ContactsList/ContactRow.jsx
```diff
@@ -15,6 +15,24 @@ import ContactEmail from './Contacts/ContactEmail'
...
+
+    const hasBreakpointsBeenChanged = Object.entries(breakpoints).some(
+      ([key, breakpoint]) => breakpoint !== nextProps.breakpoints[key]
+    )
```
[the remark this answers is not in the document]
JF-Cozy: Yes it is, but in this way we also avoid rendering at each pixel when resizing the window, which I find makes the browser  more reactive, specially with multiple contacts in the list.

src/components/ContactsList/ContactRow.jsx
JF-Cozy: No, no bug, I'm using lodash here because `contact.cozyMetadata` might be undefined so I can't just do `cozyMetadata.updatedAt !== contact.cozyMetadata.updatedAt`, that's all

src/components/ContactsList/ContactRow.jsx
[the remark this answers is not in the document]
y-lohse: I see, but the tradeoff is that `shouldComponentUpdate` is slightly slower even when the pixels don't move, for example after editing a contact. IMO resizing the window pixel by pixel is extremely rare in real life, so optimising for that over other re-renders might not be worth it, but I don't have any measures to back it up 😁. It's not a blocker for me.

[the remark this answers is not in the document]
JF-Cozy: Actually tests fails with just `!==` comparison

y-lohse (approved):
All good, thanks!

PR #680, opened 2020-09-08
fix: ContactForm no longer show old contact after submit

JF-Cozy:
At the time of the creation/editing of a contact, when a field was modified, during the validation of the form, the old data was re-displayed (because of re-rendering the component with old data)

### DIFF
--- a/src/components/Modals/ContactFormModal.jsx
+++ b/src/components/Modals/ContactFormModal.jsx
@@ -23,12 +23,31 @@ const ContactFormModal = ({
   t
 }) => {
   const [isFormBeingSubmitted, setIsFormBeingSubmitted] = useState(false)
+  const [contactInForm, setContactInForm] = useState(contact)
 
-  const submitForm = event => {
-    setIsFormBeingSubmitted(true)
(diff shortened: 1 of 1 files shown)


y-lohse:
Quick question:

> the old data was re-displayed (because of re-rendering the component with old data)

Why isn't it possible to re-render the component with data that's up to date ?

y-lohse (commented):
The fix looks good to me but I find the code is becoming hard to follow between ContactForm.onSubmit, submitForm and setFormPending. I think maybe grouping things in a single `handleSubmit` function would make the flow clearer?

JF-Cozy:
> Quick question:
> 
> > the old data was re-displayed (because of re-rendering the component with old data)
> 
> Why isn't it possible to re-render the component with data that's up to date ?

Because the data comes from the properties, and if the rendering is done before the form is submitted (which was the case before) the data coming from the properties is not yet updated.

JF-Cozy:
We can't have only one `handleSubmit` but after discussion, here's the new clearer code :)

y-lohse (approved):
Thanks, much clearer 👍

PR #681, opened 2020-09-08
Improvements

ptbrowne:
Some commits made from the update-cozy-scripts-to-4.

I prefer to split them and merge them as soon as possible
since I have some trouble upgrading cozy-scripts up
to 4.


package.json
```diff
@@ -38,7 +38,6 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-preset-cozy-app": "1.9.1",
     "bundlesize": "0.18.0",
-    "cozy-ach": "1.30.0",
```
JF-Cozy, line 43: For my knowledge, why cozy-ach should be installed globally ? Shouldn't we put a note in the doc about this?

ptbrowne: Because it is useful across repositories.

> Shouldn't we put a note in the doc about this?

It is mentioned here : https://github.com/cozy/ach#install

JF-Cozy: Sorry I meant a note in Contacts app documentation ?

ptbrowne: cozy-ach is a tool used globally across cozy, I think we should not put documentation for every tool that we use inside each of our repo, I think the documentation for cozy-ach is easily accessible from https://docs.cozy.io.

ptbrowne: But you're right, a little mention does not hurt. Done in cd5f6911fc985854bea5982c83b989a1dafbe166.

JF-Cozy (approved):
👍

PR #683, opened 2020-09-09
fix: List doesn't show up if service not installed

JF-Cozy:
If the service was not installed, the list was not displayed and remained stuck on the spinner because of the function to retrieve the service that returned an error.

fix https://github.com/cozy/cozy-contacts/issues/682

### DIFF
--- a/src/helpers/fetches.js
+++ b/src/helpers/fetches.js
@@ -1,5 +1,6 @@
+import isEqual from 'lodash/isEqual'
 import log from 'cozy-logger'
-import { isEqual } from 'lodash'
+
 import { DOCTYPE_CONTACTS } from './doctypes'
 import { updateIndexFullNameAndDisplayName } from './contacts'
 
@@ -70,6 +71,7 @@ export const fetchNormalizedServiceByName = async (client, serviceName) => {
 
     return normalizedTrigger.data
   } catch (e) {
-    throw new Error(`Can't find ${serviceName} : ${e}`)
+    log('error', `Can't find ${serviceName} : ${e}`)
+    return null
   }
 }
--- a/src/helpers/fetches.spec.js
+++ b/src/helpers/fetches.spec.js
@@ -61,4 +61,47 @@ describe('fetchNormalizedServiceByName', () => {
       expected
     )
   })
+
+  it('should return null if no trigger found', async () => {
+    const trigger = {
+      data: []
+    }
+
+    client.query = jest.fn().mockReturnValueOnce(trigger)
+
+    expect(await fetchNormalizedServiceByName(client, 'serviceName')).toEqual(
+      null
+    )
+  })
+
+  it('should return null if query returns undefined', async () => {
+    const trigger = undefined
+
+    client.query = jest.fn().mockReturnValueOnce(trigger)
+
+    expect(await fetchNormalizedServiceByName(client, 'serviceName')).toEqual(
+      null
+    )
+  })
+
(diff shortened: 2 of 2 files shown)


src/helpers/fetches.js
```diff
@@ -70,6 +71,7 @@ export const fetchNormalizedServiceByName = async (client, serviceName) => {
 
     return normalizedTrigger.data
   } catch (e) {
-    throw new Error(`Can't find ${serviceName} : ${e}`)
+    log('error', `Can't find ${serviceName} : ${e}`)
+    return null
```
JF-Cozy, line 75: Is it ok to return null here rather than an error (knowing that this is a function that is supposed to return the service) or is it better to return an error, and put conditions in the function that return null at the right time?

src/components/App.jsx
```diff
@@ -41,14 +41,23 @@ const ContactsApp = props => {
...
+        const serviceToLaunch = await fetchNormalizedServiceByName(
+          client,
+          'keepIndexFullNameAndDisplayNameUpToDate'
+        )
+
+        setServiceToLaunch(serviceToLaunch)
+        setHasServiceBeenLaunched(
+          get(serviceToLaunch, 'current_state.last_success', '').length > 0
+        )
+      } catch (e) {
```
JF-Cozy, line 54: If fetchNormalizedServiceByName returns null in case of a problem, is the catch here still relevant?

ptbrowne: If the wrapped code will not throw, the catch is not relevant anymore.

src/helpers/fetches.js
ptbrowne: It depends on the convention of your service. If it is expected that the service will sometimes not be installed, it makes to sense that a `null` expresses the fact that the service has not been found. Removing the need of a try/catch for the caller makes it easier to write the calling code, but the calling code won't be able to discriminate between a legitimate error (the service has not been found) and a (maybe) unlegitimate one (no internet). Since you are designing the function, it's you that has to make the call on what is the most convenient for the caller, and which trade-offs you are willing to make.

ptbrowne:
I think that it's better now that the list is shown even if the service is not installed.

However, the state of the code is still not satisfactory to me. There's a lot of business logic in App.jsx and I think it should not be the role of App to do this. It should be decoupled in another class or function, maybe as part of the redux store.

Moreover, I find the hook logic particularly difficult to grasp here, and I would feel safer if it
was tested (all the states possible). I think that maybe hooks are not the right tool here.

JF-Cozy:
I agree, I'll move this logic at least in an other component

ptbrowne:
I am surprised not to have any change in App.jsx. Ithink that in App.jsx, we should check if the service to launch is null and return early if it's the case.

https://github.com/cozy/cozy-contacts/blob/fix/serviceNotInstalled/src/components/App.jsx#L44

JF-Cozy:
> I am surprised not to have any change in App.jsx. Ithink that in App.jsx, we should check if the service to launch is null and return early if it's the case.
> 
> https://github.com/cozy/cozy-contacts/blob/fix/serviceNotInstalled/src/components/App.jsx#L44

I thought it wasn't necessary, because in any case you have to change the state of `hasServiceBeenLaunched' to true|false and not leave it to null. So you have to do a setState anyway. Returning earlier only saves us `setServiceToLaunch' which won't change much for React because it already does a state change.

ptbrowne:
OK I see, thanks.

ptbrowne (approved):

PR #693, opened 2020-09-21
fix: Remove accents in categorization

JF-Cozy:
Accents are removed when creating categories, to regroup contacts for example with "A" and "À" first letter

- Add test for null Indexes

According to https://medium.com/javascript-in-plain-english/not-so-obvious-removal-of-diacritics-in-javascript-explained-and-done-right-52f4aeb3c85 we chose to use `remove-accents`

### DIFF
--- a/src/helpers/contactList.spec.js
+++ b/src/helpers/contactList.spec.js
@@ -10,13 +10,17 @@ describe('categorizeContacts', () => {
       { name: 'Xavier', indexes: { byFamilyNameGivenNameEmailCozyUrl: 'X' } },
       { name: 'Zorro', indexes: { byFamilyNameGivenNameEmailCozyUrl: 'Z' } },
       { name: '', indexes: { byFamilyNameGivenNameEmailCozyUrl: '' } },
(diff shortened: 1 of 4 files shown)


src/helpers/contactList.spec.js
```diff
@@ -10,7 +10,8 @@ describe('categorizeContacts', () => {
       { name: 'Xavier', indexes: { byFamilyNameGivenNameEmailCozyUrl: 'X' } },
       { name: 'Zorro', indexes: { byFamilyNameGivenNameEmailCozyUrl: 'Z' } },
       { name: '', indexes: { byFamilyNameGivenNameEmailCozyUrl: '' } },
-      { name: {}, indexes: { byFamilyNameGivenNameEmailCozyUrl: {} } }
+      { name: {}, indexes: { byFamilyNameGivenNameEmailCozyUrl: {} } },
+      { name: 'John', indexes: { byFamilyNameGivenNameEmailCozyUrl: null } }
```
Crash--, line 14: You should add a test with À or È

Crash-- (approved):

PR #696, opened 2020-09-21
feat: Add import action menu

JF-Cozy:
- extract from StoreButton and use function to get store url
- add dropdown button and action menu to import contacts from vcard or store

### DIFF
--- a/src/components/Components/ImportDropdown.jsx
+++ b/src/components/Components/ImportDropdown.jsx
@@ -1,0 +1,72 @@
+import React, { useState, useRef } from 'react'
+import { PropTypes } from 'prop-types'
+
+import { useClient } from 'cozy-client'
+import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+import ActionMenu, { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu'
+import Icon from 'cozy-ui/transpiled/react/Icon'
+import Button from 'cozy-ui/transpiled/react/Button'
+import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
+
+import withModal from '../HOCs/withModal'
+import ContactImportationModal from '../ContactImportationModal'
+import { getFilteredStoreUrl } from '../../helpers/store'
+
+const ImportDropdown = ({ showModal, hideModal }) => {
+  const { t } = useI18n()
+  const client = useClient()
+  const anchorRef = useRef()
+  const [menuDisplayed, setMenuDisplayed] = useState(false)
+  const showMenu = () => setMenuDisplayed(true)
+  const hideMenu = () => setMenuDisplayed(false)
+
+  const showContactImportationModal = () => {
+    showModal(<ContactImportationModal closeAction={hideModal} />)
+  }
+
+  const goToStore = () => {
+    window.location = getFilteredStoreUrl(client)
+  }
+
+  return (
+    <>
+      <span ref={anchorRef}>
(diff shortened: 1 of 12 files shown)


src/components/Components/StoreButton.jsx
```diff
@@ -4,22 +4,8 @@ import { useClient } from 'cozy-client'
 import { useI18n } from 'cozy-ui/transpiled/react/I18n'
 import { ButtonLink } from 'cozy-ui/transpiled/react/Button'
 import AppIcon from 'cozy-ui/transpiled/react/AppIcon'
-import { generateUniversalLink } from 'cozy-ui/transpiled/react/AppLinker'
```
ptbrowne: (This comment is not about the above line, it's about the file). Why is there "components" twice in the path of the file ? src/components/StoreButton.jsx is not sufficient ?

src/components/Components/ImportDropdown.jsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState, useRef } from 'react'
```
ptbrowne, line 1: same question for this component, why not "src/components/ImportDropdown" ?

src/components/Components/ImportDropdown.jsx
```diff
@@ -0,0 +1,72 @@
...
+const ImportDropdown = ({ showModal, hideModal }) => {
+  const { t } = useI18n()
+  const client = useClient()
+  const anchorRef = useRef()
+  const [menuDisplayed, setMenuDisplayed] = useState(false)
+  const showMenu = () => setMenuDisplayed(true)
+  const hideMenu = () => setMenuDisplayed(false)
+
+  const showContactImportationModal = () => {
+    showModal(<ContactImportationModal closeAction={hideModal} />)
```
ptbrowne, line 24: This is a bit dangerous to do that since the modal is no longer in the React tree. It will not-render in case of change of props etc... It should not be important in this particular case since only `hideModal` is passed and it should not changed.

src/components/Components/StoreButton.jsx
JF-Cozy: `/components` contains all React components, and `/components/Components` only contains small reusable generic components such as buttons, spinner... do we have a common name to use instead of `Components` ?

src/components/Components/ImportDropdown.jsx
JF-Cozy: Okay, I'll keep that in mind. For the modals, the whole app is made like this, maybe this will be the opportunity to revisit it by migrating to `<Dialog>`.

src/components/Components/ImportDropdown.jsx
JF-Cozy: `ImportDropdown` is a small reusable generic component

src/components/Components/StoreButton.jsx
ptbrowne: >  Do we have a common name to use instead of Components ?

`/components/common` ? In Banks, page like components are grouped by functionality in "ducks" folders, and there is also a `/components` folder for common/generic components.

I personally do not care much about folder location since I navigate mostly by jumping from file to file with a search palette, but here I think that `components/Components` is not the best we can do.

ptbrowne (approved):

src/components/Components/StoreButton.jsx
JF-Cozy: I will create an other PR to not pollute this one.

PR #697, opened 2020-09-22
refactor: Rename generic component folder

JF-Cozy:
Change folder's name for generic reusable component
From Components to Common, to be clearer


ptbrowne:
Nit: OK, I think it's better to have lowercase names for everything that is not a component.

ptbrowne (approved):

Crash-- (approved):

PR #728, opened 2020-10-19
Change title on contacts empty list if filtering

JF-Cozy:
- refactor of contacts empty list
- remove useless hasConnector condition
- add new title


src/components/ContactsList/ContactsEmptyList.jsx
```diff
@@ -1,8 +1,8 @@
-import React from 'react'
-import flow from 'lodash/flow'
+import React, { useContext } from 'react'
+import { PropTypes } from 'prop-types'
```
ptbrowne, line 2: ? Shouldn't this be `import PropTypes from 'prop-types'` ?

ptbrowne (approved):

PR #731, opened 2020-10-24
feat: Navigate by category

xseignard:
With a list of the given categories in the toolbar, add the ability to scroll to the given category in the categorized contact list


xseignard:
Took into account the review and remarks of @ptbrowne

PR #763, opened 2020-12-01
feat: Use partial index for performancesPartial index

paultranvan:
Since https://github.com/cozy/cozy-client/pull/838/files , we can now use the CouchDB's partial index. In contacts, we have 2 main queries that
- Retrieve docs *with* the `indexes.byFamilyNameGivenNameEmailCozyUrl` attribute , and not trashed
- Retrieve docs *without* the `indexes.byFamilyNameGivenNameEmailCozyUrl` attribute, and not trashed

Those 2 queries can take benefit from a partial index to filter documents that do not validate those conditions. 
In local, to retrieve 1000 contacts the 2 queries used to take ~1800ms each to be performed, with all contacts having the   `indexes.byFamilyNameGivenNameEmailCozyUrl` attribute. By using partial index, the first query takes ~1200ms and the second ~15ms (as there is no contact to fetch)


Crash-- (approved):
🎉

JF-Cozy (approved):

PR #764, opened 2020-12-02
fix: Topbar height with just few contacts

JF-Cozy:
Topbar height is too high when only few contacts

will be merged as soon as CI will be green


Crash-- (approved):
don't like the "merging when is green". Next "merging even without CI"? ^^'

PR #767, opened 2020-12-16
fix: Remove partial indexes on queries

JF-Cozy:
see https://github.com/cozy/cozy-client/issues/845


ptbrowne:
Why not fixing directly the issue in cozy-client ?

JF-Cozy:
will be fixed by https://github.com/cozy/cozy-client/pull/846

PR #783, opened 2021-01-28
refactor: Add appName as prop in <App />

JF-Cozy:
Use cozy-client to get application name

### DIFF
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -2,6 +2,7 @@
 import React, { useEffect } from 'react'
 import flow from 'lodash/flow'
 
+import { useClient } from 'cozy-client'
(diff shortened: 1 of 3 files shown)


src/components/App.jsx
```diff
@@ -34,7 +34,7 @@ const ContactsApp = props => {
         {isMobile && (
           <BarCenter>
             <MuiCozyTheme>
-              <Typography variant="h4">Contacts</Typography>
+              <Typography variant="h4">{appName}</Typography>
```
ptbrowne, line 37: I think the appName is already present in cozy-client, and that instead of being threaded through props, it should be accessed via cozy-client.

ptbrowne (approved):

PR #822, opened 2021-07-09
fix: Add realtime to contacts

Crash--:
When we arrive on the app, we can launch
the service to keep the index up to date.
By doing so, the service can upgrate the
contact's documents.

In that case, if we try to edit / delete a contact
that have been edited by the service, we will get
a 409 error since the app doesn't have the right
version of the document.

To fix this issue, let's connect to the realtime
to get the latest revision if a document is changed
by something.

### DIFF
--- a/src/targets/browser/setupApp.jsx
+++ b/src/targets/browser/setupApp.jsx
@@ -1,4 +1,5 @@
 import memoize from 'lodash/memoize'
+import { RealtimePlugin } from 'cozy-realtime'
 
 import configureStore from 'store/configureStore'
 import { initTranslation } from 'cozy-ui/transpiled/react/I18n'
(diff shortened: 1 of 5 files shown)


Merkur39 (approved):

src/targets/browser/setupApp.jsx
```diff
@@ -14,6 +15,8 @@ const setupApp = memoize(() => {
   const { lang, appName } = getValues(JSON.parse(root.dataset.cozy))
   const polyglot = initTranslation(lang, lang => require(`locales/${lang}`))
   const client = getClient()
+  client.registerPlugin(RealtimePlugin)
```
Ldoppea, line 18: Only two lines of code to add realtime sync? What kind of sorcery is this? 😱

Crash--: You can create an entry for a future PDC session. I will explain how it works :)

JF-Cozy (approved):

PR #893, opened 2022-11-08
Fix duplicate contacts 

zatteo:
fix: Remove duplicate contacts by upgrading cozy-client to fix partialIndex with CouchDB 3

refactor: Use Q instead of client.find to create query and add partialIndex
    
Following the update to cozy-client 34, new warnings appeared to remove client.find and add partialIndex.

Teamwork with @cballevre and @paultranvan

### DIFF
--- a/src/thunks/cleanTrashedGroups.js
+++ b/src/thunks/cleanTrashedGroups.js
@@ -1,8 +1,11 @@
+import { Q } from 'cozy-client'
 import { DOCTYPE_CONTACT_GROUPS, DOCTYPE_CONTACTS } from '../helpers/doctypes'
 
 const cleanTrashedGroups = () => async (dispatch, getState, { client }) => {
   const { data: trashedGroups } = await client.query(
-    client.find(DOCTYPE_CONTACT_GROUPS).where({ trashed: true })
+    Q(DOCTYPE_CONTACT_GROUPS)
+      .where({ trashed: true })
+      .indexFields(['trashed'])
   )
   //eslint-disable-next-line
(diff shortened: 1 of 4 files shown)


src/thunks/cleanTrashedGroups.js
```diff
@@ -1,8 +1,11 @@
+import { Q } from 'cozy-client'
 import { DOCTYPE_CONTACT_GROUPS, DOCTYPE_CONTACTS } from '../helpers/doctypes'
 
 const cleanTrashedGroups = () => async (dispatch, getState, { client }) => {
   const { data: trashedGroups } = await client.query(
-    client.find(DOCTYPE_CONTACT_GROUPS).where({ trashed: true })
+    Q(DOCTYPE_CONTACT_GROUPS)
+      .where({ trashed: true })
```
paultranvan, line 7: This should be in a partial index: as it is a static predicate, there is no gain to use a selector

src/thunks/cleanTrashedGroups.js
```diff
@@ -33,6 +34,7 @@ const removeGroupFromAllContacts = async (client, groupId) => {
           }
         }
       })
+      .indexFields(['relationships.groups.data'])
   )
```
paultranvan, line 39: This is unrelated to the PR, but those queries should probably be moved in the dedicated file `src/helper/queries.js`

src/thunks/cleanTrashedGroups.js
zatteo: I agree, a lot of refactoring/optimization work is possible.

PR #894, opened 2022-11-10
fix: Groups took time to delete

zatteo:
In https://github.com/cozy/cozy-contacts/pull/893, I changed a _.where_ to a _.partialIndex_, and now when you click on the bin to delete a group, it disappears after 3s instead of 0s.

Adding a  _.where_ before the _.partialIndex_ fixes the problem and should not impact performance, but 
1. Do you know why ?
2. Is it a good fix ?

@paultranvan @Crash--

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "classnames": "2.3.1",
     "cozy-app-publish": "^0.29.0",
     "cozy-bar": "7.17.9",
-    "cozy-client": "^34.0.1",
+    "cozy-client": "^34.1.3",
     "cozy-device-helper": "^2.1.0",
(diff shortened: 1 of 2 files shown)


paultranvan:
> Is it a good fix ?
No, this kind of query will force a full-scan of a the database, resulting in very bad performance if there are a lot of contacts

> Do you know why ?
Nope, let's find out!

Crash-- (approved):
(you can merge it if you want before the Cozy Client upgrade. If you do so, let's add a link to your PR. If not, then request a new review after upgraded cozy-client)

Crash-- (approved):

PR #909, opened 2023-01-12
fix: use AppLinker in contact import

acezard:
It replaces legacy window.location and also handles native app.
Though the redirection doesn't work in local dev, because going
from http to https. I don't know if we want to handle that?

```
### 🔧 Bug

* Replace import contact action with an Applinker for better multi-platform behaviour
```

### DIFF
--- a/src/components/Common/ImportDropdown.jsx
+++ b/src/components/Common/ImportDropdown.jsx
@@ -1,14 +1,17 @@
 import React, { useState, useRef } from 'react'
-
-import { useClient } from 'cozy-client'
 import { useNavigate } from 'react-router-dom'
-import { useI18n } from 'cozy-ui/transpiled/react/I18n'
+
 import ActionMenu, { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu'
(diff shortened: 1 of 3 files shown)


Crash--:
> Though the redirection doesn't work in local dev, because going
from http to https. 

Are you sure this is because of that? And not because the link is something like: `https://link.mycozy.cloud/...?redirect=http://... ? ` 

If this is because of that, then we should not use https://github.com/cozy/cozy-contacts/blob/e9c06214db51da16606a93640ccaaeb8bf295a0d/src/helpers/store.js#L11 here but generateWebLink instead?

acezard:
> > Though the redirection doesn't work in local dev, because going
> > from http to https.
> 
> Are you sure this is because of that? And not because the link is something like: `https://link.mycozy.cloud/...?redirect=http://... ? `
> 
> If this is because of that, then we should not use
> 
> https://github.com/cozy/cozy-contacts/blob/e9c06214db51da16606a93640ccaaeb8bf295a0d/src/helpers/store.js#L11
> 
> here but generateWebLink instead?

I think you're right, putting back to draft while I check on that

Crash-- (approved):

PR #948, opened 2023-09-01
Feat/Change the selected group when removing it

Merkur39:
```
### ✨ Features

* If the currently selected group is deleted, default filter is set
* Improved wording when deleting group
```


Crash-- (approved):
Nice!

PR #953, opened 2023-12-22
chore: Tests are started two times

zatteo:


cballevre:
Could you take the opportunity to add this code so that builds on branches are triggered only for master and release tags?
```
branches:
  only:
    - master
     # detect also tag like x.y.z or x.y.z-beta.n as travis consider them to be branches
    - /^\d+\.\d+\.\d+(\-beta.\d+)?$/
```

cballevre (approved):

PR #955, opened 2024-01-15
feat: Show mySelf always at the top of the list

Merkur39:
Performance tests were carried out on the parent function, the before/after results are identical (between 164-177ms with 2200 contacts)

```
### ✨ Features

* Show mySelf always at the top of the list
```

### DIFF
--- a/src/helpers/contacts.js
+++ b/src/helpers/contacts.js
@@ -2,6 +2,8 @@ import sortBy from 'lodash/sortBy'
 
 import { models, HasMany } from 'cozy-client'
 
+import { filterWithRemaining } from './filterWithRemaining'
+
 const { makeFullname, makeDefaultSortIndexValue, makeDisplayName } =
   models.contact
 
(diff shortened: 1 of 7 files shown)


src/helpers/contacts.js
```diff
@@ -125,9 +127,15 @@ export const harmonizeAndSortByFamilyNameGivenNameEmailCozyUrl = (
     updateIndexFullNameAndDisplayName(contact)
   )
   const concatedContacts = contactsWithIndexes.concat(updatedContacts)
-  const sortedContacts = sortBy(concatedContacts, [
+  const {
+    itemsFound: [mySelf],
```
zatteo, line 131: Sure that there is only one `mySelf` ?

src/helpers/filterWithRemaining.js
```diff
@@ -0,0 +1,28 @@
...
+ * The function is called with the following arguments:
+ * - item: The current element in the array.
+ * - index: The index (position) of the current element in the array.
+ * @returns {FilterWithRemainingResponse} The items found and remaining.
+ */
+export const filterWithRemaining = (array, fn) => {
+  const [itemsFound, remainingItems] = [[], []]
+
+  array.forEach((arr, index) => {
+    const currentItem = arr?.model ?? arr
```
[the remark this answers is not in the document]
Merkur39: This function was taken from another app.
Indeed, this is useless here

zatteo (approved):

src/helpers/contacts.js
Merkur39: Yes "mySelf" here corresponds to `contact.me`, which is an attribute defined only on the Cozy user.

PR #956, opened 2024-01-18
fix: Correct order of contacts

Merkur39:
We did a manual sort and put "me" first in contacts list if the service was never launched.
If the service has already been launched, we retrieve the contacts such as (because it is the query which does the asc sorting), but in this case we must also put "me" first.


JF-Cozy (approved):
concernant les tests je comprends qu'ils n'était pas bon ? 🤔 on aurait peut-être pu fixer ça d'abord, puis appliquer la feature...

PR #1005, opened 2024-07-24
Fix some id issue on form fields

JF-Cozy:
```
### 🐛 Bug Fixes

* Id on contact form fields wasn't unique
* Custom labels were lost when deleting a field in the contact form arrays
```

### DIFF
--- a/src/components/ContactCard/ContactFieldInput.jsx
+++ b/src/components/ContactCard/ContactFieldInput.jsx
@@ -1,3 +1,4 @@
+import uniqueId from 'lodash/uniqueId'
 import PropTypes from 'prop-types'
 import React, { useState } from 'react'
 import { Field } from 'react-final-form'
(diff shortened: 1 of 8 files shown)


src/components/ContactCard/ContactFieldInput.jsx
```diff
@@ -42,6 +43,7 @@ const ContactFieldInput = ({
     <div className="contact-form-field__wrapper u-flex u-flex-column-s">
       <Field
         {...propsUpdated}
+        id={uniqueId('field_')}
```
Ldoppea, line 46: My understanding is that this will call `uniqueId()` on each `<ContactFieldInput>` rerender and so `id` will change too often

Suggestion: you can move it into a state like this and so id won't change during the component's life:
```jsx
const [id] = useState(uniqueId('field_'))

// ...

<Field id={id} ... />
```

Ldoppea (approved):

src/components/ContactCard/ContactFieldInput.jsx
JF-Cozy: bien vu merci, je corriger avant de merger 👍

PR #1018, opened 2024-11-07
fix: Creation of a contact when selecting a linked contact

Merkur39:
When we wanted to create a new contact from the linked contact addition,
the form component was rendered because the creation of the new contact triggers useQueries higher up the React tree.

This re-rendering caused the form to lose any changes it had already made, such as adding the new linked contact.

### DIFF
--- a/src/components/ContactCard/ContactForm/index.spec.jsx
+++ b/src/components/ContactCard/ContactForm/index.spec.jsx
@@ -3,7 +3,7 @@ import React from 'react'
 
 import { createMockClient } from 'cozy-client'
 
-import ContactForm from './index'
+import ContactForm, { isSameContactProp } from './index'
 import { johnDoeContact as contact } from '../../../helpers/testData'
 import AppLike from '../../../tests/Applike'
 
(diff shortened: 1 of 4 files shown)


src/components/ContactCard/ContactForm/index.spec.jsx
```diff
@@ -215,4 +215,76 @@ describe('ContactForm', () => {
...
+
+    it('should return false if both contacts have no "related" relationships', () => {
+      const prevProps = {
+        contact: {}
+      }
+      const nextProps = {
+        contact: {}
+      }
+      expect(isSameContactProp(prevProps, nextProps)).toBe(false)
+    })
```
zatteo, line 288: I do not understand why we check only related things. It is because the only rerender that annoy us are when we created a related contact ?

Merkur39: Returns are only due to the creation or update of contacts.
This can only happen directly by adding a linked contact.

zatteo (approved):

PR #1020, opened 2024-11-15
feat: Synchronize relation between contacts

Merkur39:
Following on from the work started on this PR https://github.com/cozy/cozy-contacts/pull/985.

The aim is to synchronize relationship changes between the contacts concerned.

### DIFF
--- a/src/connections/allContacts.js
+++ b/src/connections/allContacts.js
@@ -4,6 +4,10 @@ import uniqWith from 'lodash/uniqWith'
 
 import minilog from 'cozy-minilog'
 
+import {
+  getRelatedRelationshipsToUpdate,
+  updateRelatedRelationships
+} from './relatedRelationships'
 import { addGroupToContact } from '../helpers/contacts'
 import { DOCTYPE_CONTACTS } from '../helpers/doctypes'
 import { hasSelectedGroup } from '../helpers/groups'
@@ -77,14 +81,14 @@ export const deleteTrashedContacts = async client => {
 /**
  * @param {Object} params
  * @param {import('cozy-client/types/CozyClient').default} params.client - CozyClient
- * @param {boolean} params.isUpdated - Is the contact updated
+ * @param {import('cozy-client/types/types').ContactDocument} params.oldContact - Is the contact updated
  * @param {Object} params.formData - Contact data
  * @param {Object} params.selectedGroup - Selected group
  * @returns {Promise<import('cozy-client/types/types').IOCozyContact>} - Contact
  */
 export const createOrUpdateContact = async ({
   client,
-  isUpdated,
+  oldContact,
   formData,
   selectedGroup
 }) => {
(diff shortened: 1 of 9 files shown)


src/locales/fr.json
```diff
@@ -93,13 +93,14 @@
...
-      "colleague": "Collègue",
-      "coWorker": "Associé(e)",
-      "agent": "Responsable",
-      "acquaintance": "Assistance"
+      "friend": "Ami",
+      "colleague": "Collègue/Collaborateur",
+      "coWorker": "Associé/Partenaire",
+      "acquaintance": "Connaissance",
+      "helper": "Assistant/Aidant",
+      "recipient": "Bénéficiaire/aidé"
```
zatteo, line 103: nit: Il y une majuscule à Assistant/Aidant mais pas à Bénéficiaire/*a*idé

src/connections/allContacts.js
```diff
@@ -93,9 +97,31 @@ export const createOrUpdateContact = async ({
...
     updatedContact = addGroupToContact(updatedContact, selectedGroup)
   }
-  return isUpdated
-    ? client.save(updatedContact)
-    : client.create(DOCTYPE_CONTACTS, updatedContact)
+
+  const relatedRelationships = getRelatedRelationshipsToUpdate(
+    oldContact,
+    updatedContact
+  )
```
zatteo, line 104: extra nit: Peut être que c'est deux lignes pourrait être juste au dessus du `if` qui les concernent pour pas mélanger ?

src/connections/relatedRelationships.js
```diff
@@ -0,0 +1,263 @@
...
+      'related',
+      originalContactId,
+      relationConverted
+    )
+
+    return updatedRelatedContact
+  } catch (error) {
+    log.error('Error updating related contact', error)
+  }
+}
```
zatteo, line 197: So `createRelatedContact` and `updateRelatedContact` are equal ? Maybe you can use only one function called `upsertRelatedContact`. But it's your choice, sometimes it's good to have the different path explicit.

zatteo:
Some question, but very pleasant to read. Looks a lot like _clean code_.

src/connections/relatedRelationships.js
Merkur39: In fact, I prefer to separate the different actions.
(Initially, I was also thinking of using another CC helper in this function).

src/connections/relatedRelationships.js
```diff
@@ -0,0 +1,263 @@
...
+            originalContactId
+          })
+        case 'DELETE':
+          return deleteRelatedContact({
+            client,
+            relation: relatedRel.relation,
+            originalContactId
+          })
+      }
+    })
```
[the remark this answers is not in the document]
Merkur39: No, there is always only one related relationship item per contact.
Creation: Adds the entire related relationship.
Delete: Deletes the entire related relationship.
Update: Updates related relationship types.

src/connections/relatedRelationships.js
Merkur39: I will update the code because it wasn't working when removing a “relationTypes”.
This is a problem on the cozy-client side, so I'm updating “manually” to avoid delaying the feature.
See issue https://github.com/cozy/cozy-client/issues/1560

PR #1024, opened 2025-03-19
Add global search

zatteo:
```
### ✨ Features

* Add global search 

### 🐛 Bug Fixes

*

### 🔧 Tech

*
```


src/components/Modals/__snapshots__/ContactInfoContent.spec.jsx.snap
```diff
@@ -168,9 +168,7 @@ exports[`ContactInfoContent should match snapshot when groups are available but
           <div>
             <div
               className="u-mb-half u-breakword u-fz-medium"
-            >
-              1998-06-03
```
JF-Cozy, line 172: c'est normal cette perte ?

src/components/AppProviders.jsx
```diff
@@ -34,21 +35,27 @@ const AppProviders = ({ store, client, lang, polyglot, children }) => {
...
-                        </SelectedGroupProvider>
-                      </ContactsDiplayedProvider>
-                    </BarProvider>
-                  </AlertProvider>
-                </BreakpointsProvider>
-              </CozyTheme>
-            </I18n>
+            <DataProxyProvider
+              options={{
+                doctypes: ['io.cozy.files', 'io.cozy.contacts', 'io.cozy.apps']
```
JF-Cozy, line 40: nit: use CONST maybe ?

src/components/AppRouter.jsx
```diff
@@ -31,6 +33,7 @@ const AppRouter = () => {
               path="group/:groupId/delete/:groupName"
               element={<GroupDeleteConfirmationModal />}
             />
+            {BarRoutes.map(BarRoute => BarRoute)}
```
JF-Cozy, line 36: Couldn't we have used a component? I think not, because Routes necessarily expects a Route as a child, doesn't it?

JF-Cozy (approved):

src/components/AppRouter.jsx
zatteo: Exactly. I tried `{...BarRoutes}` but it does not work on contacts (it worked on drive 🤔). I did not take the time to understand.

src/components/Modals/__snapshots__/ContactInfoContent.spec.jsx.snap
zatteo: Très bien vu, ce n'était pas normal j'avais mal géré le BC de date-fns 👍

PR #1042, opened 2025-06-05
Add hook to tell the parent to trigger backdrop

JF-Cozy:
the entire app can be used in an iframe, as in cozy-admin for example. In this case, when a modal opens, there's a backdrop behind it, but it's constrained to the iframe. To extend this backdrop to the entire app, the iframe sends a `requestBackdropOpen` command received by the parent, triggering the opening of a global backdrop.

relative to https://github.com/cozy/cozy-admin/pull/8

### DIFF
--- a/src/components/Hooks/useIntentBackdrop.jsx
+++ b/src/components/Hooks/useIntentBackdrop.jsx
@@ -1,0 +1,26 @@
+import { useEffect } from 'react'
+import { useMatch } from 'react-router-dom'
+
+export const useIntentBackdrop = () => {
+  const matchNew = useMatch('/new')
+  const matchImport = useMatch('/import')
+  const matchContact = useMatch('/:contactId')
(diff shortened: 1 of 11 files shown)


src/components/Hooks/useIntentBackdrop.jsx
```diff
@@ -0,0 +1,26 @@
...
+  const matchContactEdit = useMatch('/:contactId/edit')
+  const matchContactDelete = useMatch('/:contactId/delete')
+  const matchGroupDelete = useMatch('/group/:groupId/delete/:groupName')
+  const isMatching =
+    !!matchNew ||
+    !!matchImport ||
+    !!matchContact ||
+    !!matchContactEdit ||
+    !!matchContactDelete ||
+    !!matchGroupDelete
```
zatteo, line 17: nit: Easier to read if !! is placed before like `  const matchNew = !!useMatch('/new')`

PR #1043, opened 2025-06-06
Migrate to rsbuild

zatteo:
Lot of tests to do, but I succeeded in building locally.

Tests
- [x] apps
- [x] services
- [x] intents


src/components/Common/Spinner.jsx
JF-Cozy, line 1: warning here, css classes will have hash now, we won't be able to override them easily... but I think it's not an issue I don't think there is override for this app, see with @poupotte maybe

JF-Cozy (approved):

PR #1045, opened 2025-06-24
feat(ui): Update topbar layout and search component design

lenhanphung:
- Swap content between topbar__left and topbar__right sections
- Implement new search component interface
- Adjust CSS styling to match Figma design specifications


zatteo (approved):
Just pass CI 👍

src/components/ContentResult.jsx
```diff
@@ -25,7 +25,10 @@ import { filterContactsBySearch, delayedSetThreshold } from '../helpers/search'
 const groupsSelectCustomStyles = {
   container: base => ({
     ...base,
-    width: '100%'
+    width: '50%',
+    '@media (max-width: 48rem)': {
```
JF-Cozy, line 29: we avoid to use hardcoded breakpoint, you can use u-classes instead on `ControlDefault` I think, or an other component styled with makeStyle perhaps. Or maybe with style from app and so css classes

JF-Cozy (approved):
nit: for next time, there is three different things in PR description, you can so make 3 different commits to give better readabilty

lenhanphung:
> nit: for next time, there is three different things in PR description, you can so make 3 different commits to give better readabilty

sure

PR #1047, opened 2025-07-03
fix: adjust store icon :bug:

lethemanh:


src/components/Common/ImportDropdown.jsx
```diff
@@ -50,7 +50,7 @@ const ImportDropdown = () => {
           <AppLinker app={{ slug: 'store' }} href={getFilteredStoreUrl(client)}>
             {({ onClick, href }) => (
               <ActionMenuItem
-                left={<AppIcon app="store" className="u-h-1 u-w-1" />}
+                left={<Icon icon={StoreIcon} className="u-h-1 u-w-1" />}
```
zatteo, line 53: @JF-Cozy does <AppIcon> mean that it is loading icon from "store" app in stable release?

JF-Cozy: https://github.com/cozy/cozy-ui/blob/master/react/AppIcon/index.jsx#L53 then https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/src/getIconURL.js#L144 so from stack (by default) or registry.

So I think we have to dig in those direction and not changing anything here

zatteo: Ok right, so it should be automatically fixed when updating `store` to stable release. I suggest we keep this PR until monday if we need it.

PR #1052, opened 2025-08-20
Refactor code to isolate the header

JF-Cozy:
The header: 

image

The goal here is to isolate the header's code to export it in cozy-ui. Then we will import it here and in the admin-panel. 

Why we move out some logic from the inner component and put them in the main component to create `onGroupCreate` sort of prop, you'll tell me? I'm glad you ask my dear! It's because in the admin-panel, we won't treat data like we do in other applications, we won't create/update/delete date on the same twake workplace but on an other one with a specific API.


zatteo:
One strange commit description : `refactor(GroupsSelect): Move code for futu`

zatteo (approved):
Clean PR 🎉

JF-Cozy:
> One strange commit description : `refactor(GroupsSelect): Move code for futu`

Oh yes good catch, I forgot to squash it with an other one (with the same name, that's why)

